PR #13, opened 2022-07-12
Node 16 support

genygilles:
## Description

NodeJS 12 security support ended 30 Apr 2022: https://endoflife.date/nodejs

This PR bumps minimal required node version to 16, which is the current latest LTS (14 will still be supported for 9 months, but 18 will become LTS in october).
In order to do so, `gulp-sass` dependency had to be upgraded. And we also updated patch versions of dependencies.

Technically, it should not break too much but as best practice, we decided to bump package version to 2 (major).

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


caarmen (approved):
LGTM

PR #18, opened 2022-10-11
Fix yarn gulp serve command

JulienBolard:
- Remove tests files watcher as 'test' task does not exist.
- Add missing cb() call to 'watch' so it can complete correctly.
- No need to call cb() directly in 'serve', pass it to bowserSync.init()

## Description

Fixes gulpfile.js to make 'yarn gulp serve' work again. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ ] I've read & comply with the [contributing guidelines](https://github.com/Genymobile/genymotion-device-web-player/blob/main/CONTRIBUTING.md)
- [ ] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.
- [ ] I have made corresponding changes to the documentation (README.md).
- [ ] I've checked my modifications for any breaking changes.

PR #19, opened 2022-12-13
System 1418/update network plugin

cmajed:
## Description

Update Network widget to handle redis messages change for android version >= 8.0

Fixes #https://genymobile.atlassian.net/browse/SYSTEM-1418

## Type of change

- [ ] Handle new redis messages for android >= 0
- [ ] Update the network widget to normalize it with Desktop one
- [ ] Move baseband section to a separated widget

## Checklist


src/plugins/Network.js
```diff
@@ -47,28 +54,69 @@ module.exports = class Network extends OverlayPlugin {
...
+        }
+        const upSpeed = values[2].split(':');
+        const downSpeed = values[3].split(':');
+        const upDelay = values[4].split(':');
+        const downDelay = values[5].split(':');
+        const upPacketLoss = values[6].split(':');
+        const downPacketLoss = values[7].split(':');
+        const dnsDelay = values[8].split(':');
+
+        if (this.androidVersion < 8) {
```
pgivel, line 109: Maybe this Android version cutoff should be extracted to a constant defined at the begining of this file ? (or elsewhere) Especially since it's used multiple times throughout this file.

src/plugins/Sim.js
```diff
@@ -0,0 +1,305 @@
...
+     * @param {boolean} basebandEnabled Whether or not baseband control is enabled.
+     */
+    constructor(instance, i18n, basebandEnabled) {
+        super(instance);
+
+        // Reference instance
+        this.instance = instance;
+
+        // Register plugin
+        this.instance.sim = this;
```
caarmen, line 24: Looks like there's a cyclical dependency. Sim has a reference to instance, and instance has a reference to sim.

Not sure if there's anyway to avoid this?

src/plugins/Network.js
```diff
@@ -540,13 +540,15 @@ module.exports = class Network extends OverlayPlugin {
...
-            console.log("Selected signalStrength:" + signalStrength.label)
+            const msgs = [];
+            msgs.push('setsignalstrength mobile ' + signalStrength.name);
+            const json = {channel: 'network_profile', messages: msgs};
+            this.instance.sendEvent(json);
         } else {
-            // hide profile details
-            this.profileDetails.classList.add('gm-hidden');
+            // TODO hide or do something?
+            console.log("Selected signalStrength not found")
```
pgivel, line 380: When you run the linter, doesn't calling the console give you a warning ? I usually add this import at the begining of the file:
```
const log = require('loglevel');
log.setDefaultLevel('debug');
```
then I log errors or debug messages through the appropriate call, like this: `log.error('my error message')`. But then again maybe using a console call is enough.

src/plugins/Network.js
```diff
@@ -35,8 +35,9 @@ module.exports = class Network extends OverlayPlugin {
         // Render components
         this.renderToolbarButton();
 
-        this.androidVersion = ""
+        this.androidVersion = "";
 
+        this.widgedRendered = false;
```
pgivel, line 40:
```suggestion
        this.widgetRendered = false;
```

src/plugins/Network.js
```diff
@@ -37,6 +37,9 @@ module.exports = class Network extends OverlayPlugin {
 
         this.androidVersion = "";
 
+        this.wifiInputStatus = true;
+        this.mobileInputStatus = true;
```
pgivel, line 38:
```suggestion
        this.wifiInputChecked = true;
        this.mobileInputChecked = true;
```
I think this could be clearer, otherwise it doesn't really mean anything to me to have a status being true or false.

src/plugins/Network.js
```diff
@@ -697,4 +705,22 @@ module.exports = class Network extends OverlayPlugin {
...
+
+    /**
+     * Update mobile signal Detail information.
+     *
+     * @param {string} detail Signal detail to update.
+     * @param {string} value  New signal detail value.
+     * @param {string} reset  If true ignore value and set "".
+     */
+     updateDetail(detail, value, reset) {
+        if(! detail) {
```
pgivel, line 527:
```suggestion
        if(!detail) {
```
code style. But this should be handled by the linter ? You can try running `yarn validate` ?

src/plugins/Network.js
```diff
@@ -47,28 +54,69 @@ module.exports = class Network extends OverlayPlugin {
...
+            this.androidVersion = version[2];
+            if (!this.widgedRendered) {
+                this.renderWidget();
+            }
+        }
+    }
+
+    handleNetworkProfile(message) {
+        const values = message.split(' ');
+        if (this.androidVersion < 8 && (values.length < 9 || values[1] === "phone")) {
```
caarmen, line 96: I have a doubt about this comparison `this.androidVersion < 8`.

If I understand right: `this.androidVersion` is a string, and we're comparing it to an int. Is that right?

Will the comparison give the expected result if we have `"11"` for an `androidVersion`?

Any way to compare api levels instead? :thinking:

pgivel (changes requested):
This is looking great ! My remarks are mostly about var naming or code style, not really deep changes.

src/plugins/Network.js
```diff
@@ -47,28 +54,69 @@ module.exports = class Network extends OverlayPlugin {
...
+        } else if (values[0] === 'parameter' && values.length > 2 && values[2].includes("android_version")) {
+            const version = values[2].match(/(android_version:)(\w+)/);
+            this.androidVersion = version[2];
+            if (!this.widgedRendered) {
+                this.renderWidget();
+            }
+        }
+    }
+
+    handleNetworkProfile(message) {
```
caarmen, line 94: I found the comment in the Sim class, with the sample format of the messages, to be quite helpful.  :pray: Would it be possible to have a similar comment here? It might make it easier to understand how the message is being parsed.

Similar for `handleSettings`.

caarmen (commented):
Did a first pass. Didn't get a chance yet to understand everything in detail.

src/plugins/Network.js
caarmen: `yarn lint --fix` will fix some of the formatting issues too :wink:

src/plugins/Sim.js
pgivel: All player plugins are already setup this way, not sure it's an issue.

caarmen: Ok! :+1:

src/plugins/util/network-profiles-ge8.js
```diff
@@ -0,0 +1,43 @@
...
+    name: 'edge',
+    label: 'EDGE',
+}, {
+    id: 2,
+    name: 'gprs',
+    label: 'GPRS',
+}, {
+    id: 1,
+    name: 'gsm',
+    label: 'GSM',
```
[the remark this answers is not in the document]
jmaneyrol69: I don't really have an opinion about this... just make it as simple as possible to avoid any confusion, imho.

src/GenymotionInstance.js
```diff
@@ -113,6 +113,22 @@ module.exports = class GenymotionInstance {
             this.callbacks[eventTag] = [];
         }
         this.callbacks[eventTag].push(callback);
+        return this.callbacks[eventTag].length - 1;
```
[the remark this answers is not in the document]
caarmen: I have a doubt about returning the callback index, in order to unregister later.

If there are a few callbacks registered, when one of them removes itself, the indices of the ones after will be shifted. If one of the later callbacks tries to unregister itself, it won't have the right index anymore.

Did I misunderstand something?

src/plugins/Network.js
```diff
@@ -114,6 +118,23 @@ module.exports = class Network extends OverlayPlugin {
         });
     }
 
+    handleSettings(message) {
+        this.instance.unregisterEventCallback('settings', this.callbackIndex)
+        const values = message.split(' ');
+        if (values[0] !== 'parameter' || values.length < 2) {
+            return;
```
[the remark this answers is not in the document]
caarmen: I'm curious, actually, why do we need to unregister the callback while we process it.  Is something in our processing going to retrigger the event and put us in a stackoverflow if we're not careful?

src/plugins/Network.js
caarmen: I think the `detail` argument will always be a defined string, no? Do we need this check `if(!detail)`?

src/plugins/Network.js
```diff
@@ -374,52 +350,73 @@ module.exports = class Network extends OverlayPlugin {
...
+    changeMobileProfile() {
+        const profile = MOBILE_PROFILES.find((elem) => elem.name === this.select.value);
+        if (profile) {
+            const msgs = [];
+            msgs.push('setprofile mobile ' + profile.name);
+            const json = {channel: 'network_profile', messages: msgs};
+            this.instance.sendEvent(json);
+        } else {
+            // TODO update profile not found
+            console.log("Selected profile not found")
```
caarmen, line 367: Will the TODOs be addressed in this PR? :thinking: (There are a few of them)

src/plugins/Sim.js
```diff
@@ -0,0 +1,305 @@
...
+        this.i18n = i18n || {};
+
+        this.fields = {};
```
caarmen, line 27: I think `this.fields` is unused.

caarmen (commented):
Did a second pass, left a couple comments/questions.

Feel free to ping me for a re-review if needed :wink:

src/GenymotionInstance.js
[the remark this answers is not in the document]
cmajed: This is no more used, I'll so remove it!

src/plugins/Network.js
```diff
@@ -32,7 +32,11 @@ module.exports = class Network extends OverlayPlugin {
 
         // Render components
         this.renderToolbarButton();
-        this.renderWidget();
+
+        this.androidVersion = ""
+
+        // Listen for settings messages: "parameter <android_version:<version>"
+        this.callbackIndex = this.instance.registerEventCallback('settings', this.handleSettings.bind(this));
```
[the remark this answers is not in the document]
cmajed: since we no more unregister this will be removed! :)

src/plugins/Network.js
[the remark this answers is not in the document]
cmajed: In the first version of this dev we had need to unregister to handle the event only once to initialize a value. It is no more needed in this version!

src/plugins/Network.js
cmajed: yarn validate
yarn run v1.22.19
$ yarn lint; yarn checkstyle
$ eslint src tests gulp
$ eslint -f checkstyle -o ./tests/reports/lint/report.xml src tests gulp
Done in 3.77s.

src/plugins/Network.js
cmajed: They are already addressed we just forget ro remove the todo statement :D

src/plugins/Network.js
```diff
@@ -470,6 +486,70 @@ module.exports = class Network extends OverlayPlugin {
...
+    }
+
+    /**
+     * Update mobile signal Detail information.
+     *
+     * @param {string} detail Signal detail to update.
+     * @param {string} value  New signal detail value.
+     * @param {string} reset  If true ignore value and set "".
+     */
+    updateDetail(detail, value, reset) {
```
caarmen, line 545: I think the `detail` argument will always be a defined string, no? Do we need this check `if(!detail)`?

src/plugins/Network.js
```diff
@@ -47,26 +60,69 @@ module.exports = class Network extends OverlayPlugin {
...
+            }
+
+            if (this.widgetRendered) {
+                this.wifiInput.checked = this.wifiInputChecked;
+                this.mobileInput.checked = this.mobileInputChecked;
+                this.updateMobileSectionStatus();
+            }
+        } else if (values[0] === 'parameter' && values.length > 2 && values[2].includes('android_version')) {
+            const version = values[2].match(/(android_version:)(\w+)/);
+            this.androidVersion = version[2];
```
caarmen, line 94: What are the possible values for `version`? Can we have `8.0` and `8.1` for oreo? or is it just the major version (`8`) which is stored?

I'm wondering because it looks like `this.androidVersion` may be a string, but we compare it to ints (the constants `ANDROID_8` and `ANDROID_10`.  I have a doubt as to how a string `"11"` would compare to an int `8`.

jmaneyrol69 (approved):

pgivel (approved):

src/plugins/Network.js
cmajed: JavaScript allow comparison between string and number :) 
https://stackoverflow.com/a/28994875/2528699

pgivel: I agree with Carmen here: I know it's possible on the technical side to compare strings & numbers, but the issue I see here is that we're specifically comparing version numbers ?

I just tested a bit in the console some variations of `"9" <= 8.1` (moving around the type and values). It seems to give good results as long as we stay under 10, since at that point alphabetic ordering can get different from numeric ordering.

It feels a bit risky to me. At the very least, forcing the conversion to number could be a good thing IMO.

pgivel: image

src/plugins/Sim.js
```diff
@@ -0,0 +1,303 @@
+'use strict';
```
JulienBolard, line 1: This is not just Sim it is baseband / RIL configuration. File should be name accordingly.

src/plugins/Network.js
```diff
@@ -47,26 +60,69 @@ module.exports = class Network extends OverlayPlugin {
...
+        }
+        const upSpeed = values[2].split(':');
+        const downSpeed = values[3].split(':');
+        const upDelay = values[4].split(':');
+        const downDelay = values[5].split(':');
+        const upPacketLoss = values[6].split(':');
+        const downPacketLoss = values[7].split(':');
+        const dnsDelay = values[8].split(':');
+
+        if (this.androidVersion < ANDROID_8) {
```
JulienBolard, line 117: Please use the capability we created for that instead of hardcoding this in the UI https://github.com/Genymobile/genymotion_private_device_genymotion/blob/genymotion-8.0.0_r11/vbox86/device.mk#L191

src/plugins/Network.js
```diff
@@ -470,6 +486,70 @@ module.exports = class Network extends OverlayPlugin {
...
+     * @param {string} profile Profile name.
+     */
+    setActiveMobileProfile(profile) {
+        if (!profile) {
+            log.error('setActiveMobileProfile: Error : provided profile is empty');
+            return;
+        }
+        const mobileProfile = MOBILE_PROFILES.find((elem) => elem.name === profile);
+        if (!mobileProfile) {
+            return;
```
JulienBolard, line 503: a log.error would make sens here

src/plugins/Network.js
```diff
@@ -470,6 +486,70 @@ module.exports = class Network extends OverlayPlugin {
...
+        }
+
+        const options = this.select.getElementsByTagName('option');
+        for (let i = 0; i < options.length; i++) {
+            const option = options[i];
+            if (option.value === mobileProfile.name) {
+                option.selected = 'selected';
```
JulienBolard, line 510: you can break  here

JulienBolard: or maybe you should unselect every one else

src/plugins/Network.js
```diff
@@ -470,6 +486,70 @@ module.exports = class Network extends OverlayPlugin {
...
+        const signalStrength = MOBILE_SIGNAL_STRENGTH.find((elem) => elem.name === strength);
+        if (!signalStrength) {
+            return;
+        }
+
+        const options = this.selectMobileSignalStrength.getElementsByTagName('option');
+        for (let i = 0; i < options.length; i++) {
+            const option = options[i];
+            if (option.value === signalStrength.name) {
+                option.selected = 'selected';
```
JulienBolard, line 533: same here

src/plugins/Sim.js
```diff
@@ -0,0 +1,303 @@
...
+
+        this.basebandEnabled = basebandEnabled;
+
+        // Render components
+        this.renderToolbarButton();
+        this.renderWidget();
+
+        // Listen for baseband messages: "<sim/network> <operator/operator_name/imsi_id/phone_number> <value>"
+        this.instance.registerEventCallback('baseband', (message) => {
+            const values = message.split(' ');
```
JulienBolard, line 35: could you create a dedicated callback function instead of defining it here?

JulienBolard (changes requested):

src/GenymotionManager.js
```diff
@@ -115,6 +117,7 @@ module.exports = class GenymotionManager {
      * @param  {boolean}            options.network                Network throttling support activated. Default: true.
      * @param  {boolean}            options.baseband               Baseband controll support activated. Default: false.
      * @param  {boolean}            options.phone                  Baseband support activated. Default: true.
+     * @param  {boolean}            options.sim                    Sim throttling support activated. Default: true.
```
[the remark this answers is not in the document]
cmajed: Ignore this. We could use already exist baseband option!

src/plugins/Network.js
cmajed: non breaking is OK

src/plugins/Sim.js
cmajed: it's old code moved but you're right it's an opportunity to do cleaner

src/GenymotionInstance.js
```diff
@@ -779,6 +779,24 @@ module.exports = class GenymotionInstance {
...
+                },
+            }, {
+                widget: this.network,
+                capability: data.message.enable5G,
+                enable: (widget) => {
+                    widget.enable5G();
+                },
+                disable: (widget) => {
+                    widget.disable5G();
+                },
```
guillaume-roche, line 799: I thought about it, and here is how we could do this in a cleaner way.

1. Add a `setCapabilities()` method in the widget super class. It will store all the capabilities in an array.
2. Add a `refresh()` method in all widgets, that will read the capabilities and update the layout (or disable the widget) accordingly.
3. Loop over the widgets, and call both methods.

This way, at most one refresh of the layouts will ever occur.

src/plugins/Network.js
```diff
@@ -63,6 +60,46 @@ module.exports = class Network extends OverlayPlugin {
...
+            this.mobileGroup.classList.add('hidden');
+            this.inputMobileSignalStrength.classList.add('hidden');
+            this.newInputs.classList.add('hidden');
+            this.inputs.classList.remove('hidden');
+        }
+    }
+
+    enable5G() {
+        if (!this.support5G) {
+            this.support5G= true;
```
guillaume-roche, line 87:
```suggestion
            this.support5G = true;
```

src/plugins/BasebandRIL.js
```diff
@@ -0,0 +1,312 @@
...
+
+            const networkOperatorMMCDiv = document.createElement('div');
+            networkOperatorMMCDiv.className = 'gm-fields';
+            const networkOperatorMMCLabel = document.createElement('label');
+            networkOperatorMMCLabel.innerHTML = 'MCC/MNC';
+            this.networkOperatorMMC = document.createElement('input');
+            this.networkOperatorMMC.className = 'gm-network-network-mmc';
+            this.networkOperatorMMC.placeholder = 'eg: 20814';
+            this.networkOperatorMMC.pattern = '[0-9]{5,6}';
+            this.networkOperatorMMC.title = 'Operator MCC/MNC';
```
JulienBolard, line 120: We must use the i18n keys to enable translation everywere

src/plugins/util/mobile-signal-strength.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+/**
+ * List of possible network signal strengths to send to the VM.
+ * Used only for android versions greater or equal to 8.
+ */
+module.exports = [{
+    id: 4,
+    name: 'great',
+    label: 'Great',
```
JulienBolard, line 10: Use i18n translation

src/plugins/BasebandRIL.js
caarmen: I had a similar thought, but in fact, this code isn't new: it's moved from `src/plugins/Network.js`.  Maybe using i18n for it can be done in a separate PR?

src/plugins/Network.js
```diff
@@ -1,8 +1,13 @@
 'use strict';
 
+const log = require('loglevel');
+log.setDefaultLevel('debug');
```
guillaume-roche, line 4: Do we need to keep this?

guillaume-roche (changes requested):

src/plugins/Network.js
```diff
@@ -161,8 +163,13 @@ module.exports = class Network extends OverlayPlugin {
         const values = message.split(' ');
 
         if (values[0] === 'if') {
-            this.wifiInput.disabled = false;
-            this.mobileInput.disabled = false;
+            if (this.wifiInput) {
```
caarmen, line 166: Is it possible to use optional chaining?


```javascript
this.wifiInput?.disabled = false
```
?

Depends on what version of javascript is supported. More info here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/GenymotionInstance.js
guillaume-roche: To do in a future rework, as this is somewhat out of scope.

src/plugins/Network.js
cmajed: we need it to log errors such:
`log.error('Selected profile not found');`

src/plugins/util/mobile-signal-strength.js
guillaume-roche: Will be done later, as this is a bit out of scope

src/plugins/Network.js
```diff
@@ -47,26 +56,188 @@ module.exports = class Network extends OverlayPlugin {
...
+    }
+
+    enable5G() {
+        const profile = MOBILE_PROFILES.at(0);
+        const option = new Option(profile.label, profile.name);
+        this.selectMobileProfile.add(option);
+    }
+
+    disable5G() {
+        // Nothing to do!
```
guillaume-roche, line 158: Why?

cmajed: because the widget is constructed by default without 5G
if capability is true we add it if its fals we have nothing to do

src/plugins/Network.js
caarmen: Actually it looks like this isn't possible after all:
https://github.com/tc39/proposal-optional-chaining#not-supported

> The following is not supported, although it has some use cases; see [Issue #18](https://github.com/tc39/proposal-optional-chaining/issues/18) for discussion:
> * optional property assignment: `a?.b = c`

src/plugins/Network.js
caarmen: I don't see any more version comparisons :+1: This point looks resolved to me!

src/plugins/Network.js
```diff
@@ -463,13 +504,79 @@ module.exports = class Network extends OverlayPlugin {
...
+
+        const options = this.selectMobileProfile.getElementsByTagName('option');
+        for (let i = 0; i < options.length; i++) {
+            const option = options[i];
+            if (option.value === mobileProfile.name) {
+                option.selected = 'selected';
+                break;
+            }
+        }
+    }
```
caarmen, line 540: Using [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) can remove a little bit of boilerplate:
```javascript
        const options = this.selectMobileProfile.getElementsByTagName('option');
        const option = options.find(option => option.value === mobileProfile.name);
        if (option) {
            option.selected = 'selected';
        } 
```

cmajed: it's an array of 7 elements

gvercoutere-geny (approved):

pgivel (approved):

README.md
```diff
@@ -383,6 +383,16 @@ Enables or disables the network widget. This widget can be used to set the netwo
 
 Enables or disables the phone widget. This widget can be used to send SMS or phone call the Android virtual device.
 
+### `Baseband`
```
caarmen, line 386: I see 2 baseband sections in this file, one is `Baseband` and one is `baseband`.  They have different descriptions, and different SaaS/PaaS compatibility. Is this correct?

cmajed: baseband section was duplicated!
I removed the second section and update the first one

src/plugins/BasebandRIL.js
```diff
@@ -0,0 +1,312 @@
...
+        const toolbars = this.instance.getChildByClass(this.instance.root, 'gm-toolbar');
+        if (!toolbars) {
+            return; // if we don't have toolbar, we can't spawn the widget
+        }
+
+        const toolbar = toolbars.children[0];
+        this.toolbarBtn = document.createElement('li');
+        this.toolbarBtnImage = document.createElement('div');
+        this.toolbarBtnImage.className = 'gm-icon-button gm-sim-button';
+        this.toolbarBtnImage.title = this.i18n.NETWORK_TITLE || 'Baseband';
```
caarmen, line 83: Is it normal that we have `NETWORK_TITLE` both in this file and in `Network.js`, for the `toolbarBtnImage.title` ?

Same question for the `title.innerHTML`.

I see in tests, in `baseband.test.js` and `network.test.js` we pass `NETWORK_TITLE: 'TEST NETWORK PLUGIN TITLE'` in both cases. 

I guess I expected something like `BASEBAND_TITLE: 'TEST BASEBAND PLUGIN TITLE'`, for the baseband plugin test, but perhaps I misunderstood something?

src/plugins/Network.js
caarmen: Actually, given the existing code above which is using for loops, it's probably better to keep the for loop here too, for consistency.

src/plugins/BasebandRIL.js
cmajed: Yep it should be replaced by BASEBAND

caarmen (approved):
I have a limited knowledge of this repository, so my understanding of the PR is not 100% complete. 

But looks like the few comments I left have been accounted for :)

PR #20, opened 2023-01-16
[GOD-2285] Camera: clearer UI about microphone injection

guillaume-roche:
## Description

Some customers do want to use microphone injection. It happens that our camera feature injects both webcam and microphone into the virtual device, but the UI is only about camera.

- Added icons to show that it enables both camera and microphone when `options.microphone` is `true`.
- Update tooltip accordingly.
- Update README: add missing `microphone` option.
- Tested locally with both values for the `microphone` option. Also checked dark and light modes.

image

## Type of change

## Checklist


JulienBolard (changes requested):
This is not available on SaaS. We need to keep the old behavior / look n feel for it

guillaume-roche:
> This is not available on SaaS. We need to keep the old behavior / look n feel for it

Ah right, I missed the `microphone`. Added a commit to take it into account. Also tested with both values for this option.

cmajed (approved):

agavignet (approved):

JulienBolard (approved):

PR #21, opened 2023-01-30
Fix network widget elements duplicated

cmajed:
This is due to the fact redis messages are received after reboot while the widget is already created and did not rendered from scratch To avoid such behaviour we add 2 boolean variable to check if the mobile throttling ans 5G enabling/disabling operation was already done or not.

## Description

Please include a summary of the change and which issue is fixed if any. Please also include relevant motivation and context (a screenshot is appreciated).

Fixes #<issue>

## Type of change

## Checklist

### DIFF
--- a/src/plugins/Network.js
+++ b/src/plugins/Network.js
@@ -37,6 +37,14 @@ module.exports = class Network extends OverlayPlugin {
         this.renderToolbarButton();
         this.renderWidget();
 
+        /*
+         * Redis message for enabling/disabling mobile throttling and 5G support
+         * could be sent without rendering the widget from srach
+         * to avoid recreation of widget elements, check this parameters before.
+         */
+        this.mobileThrottlingManaged = false;
+        this.network5Gmanaged = false;
+
         this.wifiInputChecked = true;
         this.mobileInputChecked = true;
 
@@ -60,6 +68,10 @@ module.exports = class Network extends OverlayPlugin {
     }
 
     enableMobileThrottling() {
+        if (this.mobileThrottlingManaged) {
+            return;
+        }
+
         this.mobilethrottling = true;
         // Add wifi checkbox
         const wifiGroup = document.createElement('div');
@@ -133,9 +145,15 @@ module.exports = class Network extends OverlayPlugin {
             });
 
         this.updateMobileSectionStatus();
+
+        this.mobileThrottlingManaged = true;
     }
 
     disableMobileThrottling() {
+        if (this.mobileThrottlingManaged) {
+            return;
+        }
+
         this.mobilethrottling = false;
 
         // Generate input rows for network profiles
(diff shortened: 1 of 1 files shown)


src/plugins/Network.js
```diff
@@ -37,6 +37,14 @@ module.exports = class Network extends OverlayPlugin {
         this.renderToolbarButton();
         this.renderWidget();
 
+        /*
+         * Redis message for enabling/disabling mobile throttling and 5G support
+         * could be sent without rendering the widget from srach
+         * to avoid recreation of widget elements, check this parameters before.
+         */
+        this.mobileThrottlingManaged = false;
+        this.network5Gmanaged = false;
```
guillaume-roche, line 46: Maybe like this, as it seems a clearer wording to me:
```suggestion
        this.mobileThrottlingConfigured = false;
        this.network5GConfigured = false;
```

guillaume-roche (commented):
Looks good, just had one comment.

c-quenin (approved):

src/plugins/Network.js
```diff
@@ -37,6 +37,14 @@ module.exports = class Network extends OverlayPlugin {
         this.renderToolbarButton();
         this.renderWidget();
 
+        /*
+         * Redis message for enabling/disabling mobile throttling and 5G support
+         * could be sent without rendering the widget from srach
+         * to avoid recreation of widget elements, check this parameters before.
```
JulienBolard, line 43: these parameters

src/plugins/Network.js
```diff
@@ -151,16 +169,28 @@ module.exports = class Network extends OverlayPlugin {
                 this.selectProfile.add(option);
             });
         this.profileInputs.appendChild(this.selectProfile);
+
+        this.mobileThrottlingConfigured = true;
     }
 
     enable5G() {
+        if (this.network5GConfigured) {
```
JulienBolard, line 177: Why don't you just check for existing option 'profile.name', and avoid the addition if it is already present?

src/plugins/Network.js
```diff
@@ -37,6 +37,14 @@ module.exports = class Network extends OverlayPlugin {
         this.renderToolbarButton();
         this.renderWidget();
 
+        /*
+         * Redis message for enabling/disabling mobile throttling and 5G support
+         * could be sent without rendering the widget from srach
+         * to avoid recreation of widget elements, check this parameters before.
+         */
+        this.mobileThrottlingConfigured = false;
```
JulienBolard, line 45: using mobileThrottlingEnabled and network5GEnabled, you would be able to have a more future proof behavior by being able to toggle them dynamically. Here you are stuck with a configuration, and not able to modify widget dynamically. (when configuration would have change on the device for example)

src/plugins/Network.js
cmajed: 5G is always preset in mobile profile list we add it to the selector only if the enable5G method is triggered.

guillaume-roche (approved):

JulienBolard (approved):
I don't like this hack, but so be it

PR #36, opened 2023-07-24
Fix unmute popup

pgivel:
## Description

Update unmute popup:
- ~~make it clickable too (since the cursor suggests this anyway)~~ in fact, make the whole window clickable
- add i18n for this popup

This could probably be merged in the main branch too.

## Type of change

## Checklist

### DIFF
--- a/src/DeviceRenderer.js
+++ b/src/DeviceRenderer.js
@@ -564,7 +564,7 @@ module.exports = class DeviceRenderer {
                     this.dispatchEvent('video', {msg: 'play automatically allowed without sound'});
                     const popup = document.createElement('div');
                     popup.classList.add('gm-click-to-unmute');
-                    popup.innerHTML = 'By default, the sound has been turned off, '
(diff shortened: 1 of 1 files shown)


JulienBolard (approved):
This needs to be tested in SaaS and PaaS as adding the listener on the window globally could be a problem.

pgivel:
> This needs to be tested in SaaS and PaaS as adding the listener on the window globally could be a problem.

You're right, I am currently testing this on the SaaS webapp (I don't have the setup for PaaS, I'll talk about this with Jérémy). However I think it is quite safe since we remove the listener as soon as it was called once.

pgivel:
@JulienBolard I'm currently still testing on PaaS, but with SaaS the new behaviour looks OK to me:
The whole UI is still usable as normal when the click-to-unmute popup is visible; clicking **anywhere** will close the popup (even on a UI element, and that UI element will behave as normal)

jparez (approved):

PR #40, opened 2023-09-06
Refactor camera & mic features into MediaManager

pgivel:
## Description

This splits the functionnal and UI usage. It also decorrelates the microphone & camera features, allowing to stream one or the other independantly.

Fixes #<issue>

## Type of change

## Checklist


src/plugins/MediaManager.js
```diff
@@ -0,0 +1,339 @@
...
+        this.instance.mediaEventsEnabled = true;
+
+        this.videoStreaming = false;
+        this.audioStreaming = false;
+
+        this.videoWidth = videoWidth;
+        this.videoHeight = videoHeight;
+        this.videoWithMicrophone = videoWithMicrophone;
+
+        this.setupPermissions();
```
skornous, line 32: is it really necessary to redo a setupPermissions ? if [l.18](https://github.com/Genymobile/genymotion-device-web-player/pull/40/files#diff-97a1a105b0f7db226929204ef391737aac32f2a5cbaf8a20278a55453fb07cddR18) passes it should already have been done so redoing it might double register the `on<Microphone/Camera>PermissionChange` listeners

pgivel: You're right, I forgot to remove this line when adding l.18 :) Thanks!

jparez (approved):

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,341 @@
...
+
+    /**
+     * Find RTCRtpSender corresponding RTCRtpTransceiver and set its direction.
+     *
+     * @param {RTCRtpSender} sender    Peer (sender).
+     * @param {string}       direction Transceiver direction.
+     */
+    setTransceiverDirection(sender, direction) {
+        // find transceiver that contains sender
+        this.instance.peerConnection.getTransceivers().forEach((transceiver) => {
```
jparez, line 196: don't know if you can get multiple times the same sender but if not you can use find / findIndex, it's less clearer but the loop stop after getting the first row which match

const arrTransceivers =  this.instance.peerConnection.getTransceivers()
const index = arrTransceivers .findIndex(transceiver => transceiver.sender === sender)
if (index > -1)  arrTransceivers[index].direction = direction

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,341 @@
...
+            this.startAudioStreaming();
+        } else {
+            this.stopAudioStreaming();
+        }
+    }
+
+    /**
+     * Initialize and start client webcam video stream.
+     */
+    startVideoStreaming() {
```
jparez, line 90: then method is oldschool and can become unreadable as it grow. startVideoStreaming won't return a promise, so if you want to wait for startVideoStreaming in the future you can't without modify startVideoStreaming

i prefere this way

```
async startVideoStreaming() {
     if (!navigator.mediaDevices) {
         return;
     }
     try{
        const mediaStream = navigator.mediaDevices.getUserMedia({
            audio: this.videoWithMicrophone,
            video: {
                width: this.videoWidth,
                height: this.videoHeight,
            },
        })
        log.debug('Client video stream ready');
        this.videoStreaming = true;
        this.localVideoStream = mediaStream;
        this.addVideoStream(mediaStream);
      } catch((err)=> {
        this.onVideoStreamError(err);
      })           
}
```

src/plugins/Camera.js
```diff
@@ -60,73 +58,9 @@ module.exports = class Camera extends OverlayPlugin {
...
-    /**
-     * Toggle local video forward forwarding.
-     * Redirect the client webcam video stream to the instance.
-     */
-    toggleStreaming() {
-        if (!this.streaming) {
-            this.startStreaming();
-        } else {
-            this.stopStreaming();
+            this.toolbarBtn.onclick = () => {
```
jparez, line 61: maybe i wrong but i think 

`this.toolbarBtn.onclick = this.instance.mediaManager.toggleVideoStreaming`

Could get the job. Here we bind an arrow fn which call a fn.

src/DeviceRenderer.js
```diff
@@ -298,8 +299,8 @@ module.exports = class DeviceRenderer {
     disconnect() {
         this.initialized = false;
 
-        if (typeof this.camera !== 'undefined' && typeof this.camera.getClientVideoStream !== 'undefined') {
-            this.removeLocalStream(this.camera.getClientVideoStream());
+        if (typeof this.mediaManager !== 'undefined') {
```
jparez, line 302: `typeof this.mediaManager !== 'undefined' && this.mediaManager.disconnect();`

Is more sexy, isn't ?

pgivel: more sexy but less readable so I'm not sure if it's better in this case :/

src/DeviceRenderer.js
```diff
@@ -40,7 +41,6 @@ module.exports = class DeviceRenderer {
 
         // Websocket
         this.webRTCWebsocket = null;
-        this.webRTCWebsocketName = 'gm-webrtc';
```
JulienBolard, line 43: It could be preferable to defined it but set it  to null

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,341 @@
...
+            return false;
+        }
+
+        navigator.permissions.query({name: 'microphone'})
+            .then((permissionObj) => {
+                log.debug(`microphone ${permissionObj.state}`);
+                permissionObj.addEventListener('change', this.onMicrophonePermissionChange.bind(this));
+            })
+            .catch((error) => {
+                log.debug('Can\'t get microphone permission object', error);
```
JulienBolard, line 48: this should `return false;`

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,341 @@
...
+            });
+
+        navigator.permissions.query({name: 'camera'})
+            .then((permissionObj) => {
+                log.debug(`camera ${permissionObj.state}`);
+                permissionObj.addEventListener('change', this.onCameraPermissionChange.bind(this));
+            })
+            .catch((error) => {
+                log.debug('Can\'t get camera permission object', error);
```
JulienBolard, line 57: same here: `return false;`

src/plugins/MediaManager.js
JulienBolard: It is really interesting to read that .then() is old school and you propose a more conventional way of writing code. I really don't understand JS dev :)

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,341 @@
...
+            });
+    }
+
+    /**
+     * Client video stream error handler.
+     *
+     * @param {Error} error Camera stream error.
+     */
+    onVideoStreamError(error) {
+        log.warn('Can\'t start client video stream', error);
```
JulienBolard, line 142: log.error instead ?

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /**
+     * Client audio stream error handler.
+     *
+     * @param {Error} error Microphone stream error.
+     */
+    onAudioStreamError(error) {
+        log.warn('Can\'t start client audio stream', error);
```
JulienBolard, line 151: log.error instead ?

src/plugins/MediaManager.js
JulienBolard: This needs to be verified as I am not sure if there was multiple instance of the same sender when I did this dev

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,341 @@
...
+            });
+            if (this.microphoneSender) {
+                this.instance.peerConnection.removeTrack(this.microphoneSender);
+            }
+            this.instance.renegotiateWebRTCConnection();
+        }
+    }
+
+    /**
+     * Remove a local video stream and stop sending it through SDP renegotiation.
```
JulienBolard, line 304: It also remove the audio track not just the video. Not sure if it is really intended now we split audio and video

JulienBolard (changes requested):

src/plugins/MediaManager.js
pgivel: I was also surprised at this but indeed, the browser support versions seems to indicate that async/await + try/catch is much newer than promises

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#browser_compatibility

src/plugins/MediaManager.js
pgivel: I should add comments to further explain that.

tests/unit/camera.test.js
```diff
@@ -16,12 +17,13 @@ describe('Camera Plugin', () => {
             navbar: true,
             power: true
         });
-        camera = new Camera(instance, false);
+        mediaManager = new MediaManager(instance);
     });
 
     describe('api', () => {
         test('exposes a high level constructor', () => {
             expect(typeof Camera).toBe('function');
```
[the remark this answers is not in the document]
pgivel: I'm not sure if this unit test is supposed to test the camera plugin (the ui mostly) or the mediamanager plugin (the functionality). I should probably split it in two tests.

src/plugins/MediaManager.js
pgivel: I admit I'm not sure what's the point of setting the direction in the first place. I guess if the transceiver is in the `receive` direction we can't send data through the stream ? But how/why would it be in the `receive` direction?

pgivel: (since I don't understand the aim here, I'm not sure how I can verify that @jparez's suggestion doesn't break stuff)

src/DeviceRenderer.js
pgivel: You're right, it'll be more legible. But first I'll need to test if setting up the webrtc socket with a `null` protocol is the same as setting it up with an `undefined` one.

pgivel: @JulienBolard setting the protocol to `null` doesn't work, the socket is initialized with a `null` protocol instead of without a specific one, and so the connexion isn't established.

src/plugins/MediaManager.js
pgivel: closing this suggestion since the existing code works already and I don't know how to check that I'm not breaking stuff by changing it

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,345 @@
...
+     *  This flag is used when we don't want to split audio&video and instead want the audio bundled with the video in the same stream.
+     *  Defaults to the microphone flag in the instance options
+     * @param {number} videoWidth Maximum video width. Defaults to 1280
+     * @param {number} videoHeight Maximum video height. Defaults to 720
+     */
+    constructor(instance, videoWithMicrophone = instance.options.microphone, videoWidth = 1280, videoHeight = 720) {
+        this.instance = instance;
+        this.instance.mediaManager = this;
+
+        if (!navigator.mediaDevices || !this.setupPermissions()) {
```
Previsou, line 20: no need to await `this.setupPermissions()` ?

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,345 @@
...
+            const mediaStream = await navigator.mediaDevices.getUserMedia({
+                audio: this.videoWithMicrophone,
+                video: {
+                    width: this.videoWidth,
+                    height: this.videoHeight,
+                },
+            });
+            log.debug('Client video stream ready');
+            this.videoStreaming = true;
+            this.localVideoStream = mediaStream;
```
Previsou, line 109: Should we set audio related fields as well if `this.videoWithMicrophone = true` ?  Or adapt toggleAudioStreaming to protected against starting an audio stream while mic is already captured?

Previsou (approved):
New to the codebase, so more a cursory review than anything else :-)

src/plugins/MediaManager.js
jparez: Yes indeed, but correct me if i'm wrong, constructors cannot be async. 
We can use this.setupPermissions.then(()=>{}) but instance props (this.instance.mediaEventsEnabled, ...) may be initialized after they are used

I guess even if setupPermissions return false, since it's return a promise so !this.setupPermissions will be always true, the test pass.

missed this one :$

jparez (approved):

src/plugins/MediaManager.js
Previsou: > Yes indeed, but correct me if i'm wrong, constructors cannot be async.
I am far from fluent in javascript, so I really don't know. Kinda makes sense though to avoid a promise on a futur object.

src/plugins/MediaManager.js
pgivel: hm is it really an issue if the user sends a video&audio stream + an audio stream ?

src/plugins/MediaManager.js
pgivel: Is it better now?

README.md
```diff
@@ -276,7 +276,8 @@ Enables or disables the fullscreen widget. This widget can be used to make the r
 - **Default:** `true`
 - **Compatibility:** `PaaS`, `SaaS`
 - **Details:**
-Enables or disables the camera widget. This widget can be used to forward local webcam to the Android virtual device.
+Enables or disables the camera widget. This widget can be used to forward local webcam video to the Android virtual device.
+By default, if the `microphone` property is also true, then the webcam audio will be used as well.
```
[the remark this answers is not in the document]
pgivel: indeed, I'll update that

src/plugins/MediaManager.js
```diff
@@ -0,0 +1,352 @@
...
+        this.videoWithMicrophone = videoWithMicrophone;
+    }
+
+    /**
+     * Tries to get the current microphone and camera permission status. If it can, binds listeners to their change
+     * @returns {boolean} True if it could find the permissions object
+     */
+    async setupPermissions() {
+        if (!navigator.permissions) {
+            return false;
```
Previsou, line 42: Shouldn't you set `this.instance.mediaEventsEnabled = false` here? In fact, this field seems totally unused

pgivel: `mediaEventsEnabled` is to be used by other classes above this manager (for example for disabling the UI), so that's why it is unused in this class.

It is set to `true` on line 26, as long as we're able to use the `mediaDevices` API. I don't want to set it to false if we can't use the `permissions` API because the feature still works, it simply won't be able to react dynamically to permission changes.

Previsou: Thanks for the explanation

Previsou (approved):

JulienBolard (approved):

PR #44, opened 2023-11-23
Media manager: Add return values in order to fail loudly

pgivel:
## Description

This PR adds return values to most functions in media manager in order to communicate with users of this class if we failed to initialise video or audio input.

## Type of change

## Checklist

### DIFF
--- a/src/plugins/MediaManager.js
+++ b/src/plugins/MediaManager.js
@@ -68,33 +68,34 @@ module.exports = class MediaManager {
     /**
      * Toggle local video forwarding.
      * Redirect the client webcam video stream to the instance.
+     * @returns {boolean} true on success, false on fail
      */
     toggleVideoStreaming() {
         if (!this.videoStreaming) {
-            this.startVideoStreaming();
-        } else {
-            this.stopVideoStreaming();
+            return this.startVideoStreaming();
         }
+        return this.stopVideoStreaming();
     }
 
     /**
      * Toggle local audio forwarding.
      * Redirect the client microphone audio stream to the instance.
+     * @returns {boolean} true on success, false on fail
      */
     toggleAudioStreaming() {
         if (!this.audioStreaming) {
-            this.startAudioStreaming();
-        } else {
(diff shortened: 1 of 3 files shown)


skornous (approved):
lgtm

JulienBolard (approved):

src/plugins/MediaManager.js
```diff
@@ -130,9 +133,10 @@ module.exports = class MediaManager {
             log.debug('Client audio stream ready');
             this.audioStreaming = true;
             this.localAudioStream = mediaStream;
-            this.addAudioStream(mediaStream);
+            return this.addAudioStream(mediaStream);
```
jparez, line 136: if addAudioStream fail the error will be uncaught. `this.instance.renegotiateWebRTCConnection();` call renegotiateWebRTCConnection wihch do `this.peerConnection.createOffer` and createOffer return a promise.

I think `addAudioStream` need to return `this.instance.renegotiateWebRTCConnection();` or a Promise.reject() and this line should be prefix by an await.

src/plugins/MediaManager.js
```diff
@@ -156,28 +160,32 @@ module.exports = class MediaManager {
...
+     * @returns {boolean} true on success, false on fail
      */
     stopVideoStreaming() {
         if (!this.videoStreaming) {
-            return;
+            return false;
         }
         log.debug('removed local video stream');
-        this.removeVideoStream(this.localVideoStream);
+        const result = this.removeVideoStream(this.localVideoStream);
```
jparez, line 170: Same thing here, This call renegotiateWebRTCConnection which call this.peerConnection.createOffer.

src/plugins/MediaManager.js
pgivel: I was mostly focused on errors related to permissions or availability of input devices but you're right I should cover this case too. Thanks !

pgivel: Is it OK for you now ?

PR #48, opened 2024-01-08
Improve Device Renderer destruction

pgivel:
## Description

This PR improves Device Renderer destruction, in hopes of it getting garbage-collected when we don't need it anymore.

Most event handlers lambda were moved to class methods, so that they could be properly removed, and most plugins had `destroy` functions added, and called in the Device Renderer `destroy`.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/src/DeviceRenderer.js
+++ b/src/DeviceRenderer.js
@@ -70,6 +70,10 @@ module.exports = class DeviceRenderer {
         document.addEventListener('click', this.clickHandlerCloseOverlay.bind(this));
     }
 
+    /**
+     * Handler for click event, responsible for closing overlay when clicking outside of it.
+     * @param {PointerEvent} event Event that triggered this handler
+     */
     clickHandlerCloseOverlay(event) {
         if (event.target.closest('.gm-overlay') === null
             && !event.target.classList.contains('gm-icon-button')
@@ -212,7 +216,7 @@ module.exports = class DeviceRenderer {
         this.webRTCWebsocket.onclose = (event) => {
             this.video.style.background = this.videoBackupStyleBackground;
             this.initialized = false;
-            log.debug('Error! Maybe your VM is not available yet? (' + event.code +') ' + event.reason);
+            log.debug('Error! Maybe your VM is not available yet? (' + event.code + ') ' + event.reason);
 
             switch (event.code) {
             case 1000:
@@ -478,8 +482,8 @@ module.exports = class DeviceRenderer {
                         popup.remove();
                         log.debug('Playing video with sound enabled has been authorized due to user click');
                     };
-                    window.addEventListener('click', addSound, {once:true});
-                    window.addEventListener('touchend', addSound, {once:true});
+                    window.addEventListener('click', addSound, {once: true});
+                    window.addEventListener('touchend', addSound, {once: true});
                 }).catch(() => {
                     log.debug('Can\'t play video, even with sound disabled');
                     this.dispatchEvent('video', {msg: 'play denied even without sound'});
@@ -495,8 +499,8 @@ module.exports = class DeviceRenderer {
                         this.dispatchEvent('video', {msg: 'play manually allowed by click'});
                         log.debug('Playing video with sound disabled has been authorized due to user click');
                     };
-                    div.addEventListener('click', allowPlay, {once:true});
-                    div.addEventListener('touchend', allowPlay, {once:true});
+                    div.addEventListener('click', allowPlay, {once: true});
+                    div.addEventListener('touchend', allowPlay, {once: true});
                 });
             });
         };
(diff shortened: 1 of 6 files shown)


src/plugins/MediaManager.js
```diff
@@ -363,4 +365,12 @@ module.exports = class MediaManager {
             this.stopVideoStreaming();
         }
     }
+
+    /**
+     * Plugin destructor, responsible for removing all callbacks & bindings so that things are garbage-collected
+     */
+    destroy() {
+        this.microphonePermissionObject?.removeEventListener('change', this.onMicrophonePermissionChange.bind(this));
```
jparez, line 373: Maybe we can keep an array with all permissions object (or all fn to call for cleanup) and iterate over ? This way if we add new listener to cleanup all we need to do it's to add the clean up fn to the array

jparez: Maybe the array can be shared accross plugin, so throught DeviceRender we can clean up all things

src/plugins/PeerConnectionStats.js
```diff
@@ -21,7 +21,7 @@ module.exports = class PeerConnectionStats {
 
         this.connection = connection;
 
-        setTimeout(() => {
+        this.timeoutID = setTimeout(() => {
```
jparez, line 24: if you wish you can write this with an async / await


```
this.timeoutID = setTimeout(async () => {
   const stats = await this.connection.getStats(null)
   ...
}, interval)
```

src/DeviceRenderer.js
```diff
@@ -67,25 +67,31 @@ module.exports = class DeviceRenderer {
...
-        });
+        document.addEventListener('click', this.clickHandlerCloseOverlay.bind(this));
+    }
+
+    /**
+     * Handler for click event, responsible for closing overlay when clicking outside of it.
+     * @param {PointerEvent} event Event that triggered this handler
+     */
+    clickHandlerCloseOverlay(event) {
+        if (!this.hasSomeParentTheClass(event.target, 'gm-overlay')
```
jparez, line 78: The comment doesn't seem to be related to the scope of this pull request, but I'm wondering if hasSomeParentTheClass couldn't be refactored to something like

```
function hasSomeParentTheClass(element, className) {
    return element.closest("." + className) !== null;
}
```
https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

src/DeviceRenderer.js
```diff
@@ -836,4 +844,18 @@ module.exports = class DeviceRenderer {
...
+        this.peerConnection?.removeEventListener('connectionstatechange', this.onConnectionStateChange);
+        this.mediaManager?.destroy();
+        this.gamepadManager?.destroy();
+        this.peerConnectionStats?.destroy();
+        delete this.peerConnectionStats;
+        delete this.video;
+        delete this.wrapper;
+        delete this.videoWrapper;
+        delete this.root;
+        delete this.stream;
```
JulienBolard, line 872: are we missing a delete this.peerConnection ?

src/DeviceRenderer.js
```diff
@@ -836,4 +851,24 @@ module.exports = class DeviceRenderer {
...
```
jparez, line 872: We need to ensure that all addEventListener instances are removed before destruction, or this may lead to a memory leak.

src/DeviceRenderer.js
pgivel: you're probably right, I'll do this right away in another PR.

src/DeviceRenderer.js
```diff
@@ -548,12 +558,7 @@ module.exports = class DeviceRenderer {
             div.innerHTML = message;
         };
 
-        this.peerConnection.onconnectionstatechange = () => {
-            log.debug('ConnectionState changed:', this.peerConnection.iceConnectionState);
-            if (this.peerConnection.iceConnectionState === 'disconnected') {
-                this.onWebRTCReady();
-            }
-        };
+        this.peerConnection.addEventListener('connectionstatechange', this.onConnectionStateChange);
```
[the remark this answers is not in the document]
pgivel: I'm not sure I understand your question correctly, but I think doing it the simpler way like on line 70 will solve it.

For the second question, I think the point of this code isn't exactly to keep state of the connection, but to reconnect if we lose it.

src/DeviceRenderer.js
pgivel: `this.peerConnection` is closed and set to null in the `disconnect` method, and I assumed that method was called since that's the case in Shadow. However I was wrong to assume this, it's simply not true here. I'll add a `disconnect` at the very begining of the `destroy` method.

src/DeviceRenderer.js
pgivel: I'm not sure what you mean here. Isn't this what I'm doing in this function?

src/plugins/MediaManager.js
pgivel: As discussed on slack, I think this is out of scope for this PR. I'll try a POC PR based on this one, just to see if this is too big of a change or if it's OK.

jparez (approved):

JulienBolard (approved):

src/plugins/MediaManager.js
pgivel: draft PR here: https://github.com/Genymobile/genymotion-device-web-player/pull/51

PR #58, opened 2024-02-06
Remove Lodash dependency

pgivel:
## Description

defaultsDeep was causing a weird bug on Shadow's side because of change-in-place behaviour. I found it clearer to use `Object.assign` from the default library.

Lodash wasn't used anywhere else, so let's remove the dependency.

### Notes

- [Object.assign makes a shallow assignment](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Object/assign). I'm in the process of checking if this doesn't break anything.
- I would have prefered to use the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), but it's only available in ES2018 and gulp is holding us back

## Type of change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
   },
   "homepage": "https://github.com/Genymobile/genymotion-device-web-player#readme",
   "dependencies": {
-    "lodash": "^4.17.21",
     "loglevel": "^1.8.0",
     "webrtc-adapter": "^7.7.1"
   },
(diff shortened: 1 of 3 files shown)


pgivel:
Closing, as this change isn't actually necessary, the similar change on Shadow's side only was enough.

PR #64, opened 2024-03-18
[PLAYER-2] bind keypress envent on root div instead of window object

jparez:
## Description
Fix bug when typing charater outside of player events were send to the instance

## Type of change

## Checklist


pgivel (approved):
LGTM. I can test on shadow player that it doesn't break stuff if you wish, although as we discussed over slack I don't think it will since the shadow player overrides this plugin.

PR #66, opened 2024-03-20
[PLAYER-7] fix issue when destroy a not fully initialized player

jparez:
## Description

When calling setupRender and destroy() in a loop destroy() destroy try to delete object not already load

see https://genymobile.atlassian.net/browse/PLAYER-7

## Type of change

### DIFF
--- a/src/DeviceRenderer.js
+++ b/src/DeviceRenderer.js
@@ -876,6 +876,14 @@ module.exports = class DeviceRenderer {
      * This method also calls recursively the destroy methods on the plugins if they exist.
      */
     destroy() {
+        /**
+         *  if the websocket is in connecting state,
+         *  we can't destroy the instance cause object is not fully initialized
+         */
+        if (this.webRTCWebsocket.readyState === 0) {
+            return;
+        }
+
         this.removeAllListeners();
         this.disconnect();
         this.peerConnectionStats?.destroy();


src/DeviceRenderer.js
```diff
@@ -876,14 +876,20 @@ module.exports = class DeviceRenderer {
...
-        delete this.video;
-        delete this.wrapper;
-        delete this.videoWrapper;
-        delete this.root;
-        delete this.stream;
+        /**
+         *  if the websocket is in connecting state,
+         *  we can't destroy the instance cause object is not fully initialized
+         */
+        if (this.webRTCWebsocket.readyState !== 0) {
```
pgivel, line 883: I think it could be a little bit cleaner to reverse the condition:
```suggestion
        /**
         *  if the websocket is in connecting state,
         *  we can't destroy the instance cause object is not fully initialized
         */
        if (this.webRTCWebsocket.readyState === 0) {
            return;
        }
```

I don't know if we could also add an event listener for when the socket is ready ? Maybe it's useless since the page would get closed anyway? I was thinking of something like this:
```suggestion
        /**
         *  if the websocket is in connecting state,
         *  we can't destroy the instance cause object is not fully initialized
         */
        if (this.webRTCWebsocket.readyState === 0) {
            this.addListener(this.webRTCWebsocket, 'open', this.destroy.bind(this), {once: true});
            return;
        }
```

pgivel (approved):
LGTM as is, but a couple suggestions for improvment.

pgivel (approved):

PR #72, opened 2024-03-28
adding loading dependency logic

jparez:
## Description

- Adding dependency loading logic for plugin to load.
- Move all default plugin to custom plugin

firefox_45JT55dBCc

firefox_WT4KqpnE1G

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


src/DeviceRendererFactory.js
```diff
@@ -210,39 +190,36 @@ module.exports = class DeviceRendererFactory {
...
-            {enabled: options.gamepad, class: GamepadManager},
-            {enabled: options.biometrics, class: FingerPrint},
-            {enabled: options.buttons, class: ButtonsEvents, params: [options.i18n, options.translateHomeKey]},
-            {enabled: options.camera || options.microphone, class: MediaManager},
         ];
 
         if (typeof instance.addCustomPlugins === 'function') {
-            instance.addCustomPlugins();
+            const customPlugins = instance.addCustomPlugins();
+            if (Array.isArray(customPlugins)){
```
pgivel, line 203: Maybe `customPlugins` could be simplified to an array, instead of having an `addCustomPlugins` method that generates an array?

src/DeviceRendererFactory.js
```diff
@@ -210,39 +190,36 @@ module.exports = class DeviceRendererFactory {
...
+                if (plugin.dependencies) {
+                    plugin.dependencies.forEach((Dep) => {
+                        if (dependenciesLoaded.indexOf(Dep.name) !== -1) {
+                            return;
+                        }
+                        new Dep(instance);
+                        dependenciesLoaded.push(Dep.name);
+                    });
+                }
                 new plugin.class(instance, ...args);
```
pgivel, line 223: I think you need to add here `dependenciesLoaded.push(plugin.class);` otherwise if you load a dependency before the plugin that requires it, then it will be loaded twice.

src/DeviceRendererFactory.js
```diff
@@ -210,39 +190,36 @@ module.exports = class DeviceRendererFactory {
...
+
+        const dependenciesLoaded = [];
         pluginInitMap.forEach((plugin) => {
             const args = plugin.params || [];
 
             if (plugin.enabled) {
+                // load dependencies
+                if (plugin.dependencies) {
+                    plugin.dependencies.forEach((Dep) => {
+                        if (dependenciesLoaded.indexOf(Dep.name) !== -1) {
```
pgivel, line 216: I don't understand where this `name` property comes from.

src/DeviceRenderer.js
```diff
@@ -83,23 +101,39 @@ module.exports = class DeviceRenderer {
...
-            {enabled: this.options.camera, class: Camera, params: [this.options.i18n]},
+            {enabled: this.options.gamepad, class: Gamepad, params: [this.options.i18n], dependencies:[GamepadManager]},
+            {enabled: this.options.fileUpload, class: FileUpload, params: [this.options.i18n]},
+            {enabled: this.options.identifiers, class: Identifiers, params: [this.options.i18n]},
+            {enabled: this.options.network, class: Network, params: [this.options.i18n]},
+            {enabled: this.options.phone, class: Phone, params: [this.options.i18n]},
+            {enabled: this.options.baseband, class: BasebandRIL, params: [this.options.i18n, this.options.baseband]},
+            {enabled: this.options.diskIO, class: IOThrottling, params: [this.options.i18n]},
+            {enabled: this.options.biometrics, class: FingerPrint},
+            {enabled: this.options.microphone, class: MediaManager},
```
pgivel, line 133: I think `MediaManager` should stay as a default plugin inside the `DeviceRendererFactory`.

src/DeviceRenderer.js
```diff
@@ -83,23 +101,39 @@ module.exports = class DeviceRenderer {
...
+                enabled: this.options.buttons,
+                class: ButtonsEvents,
+                params: [this.options.i18n, this.options.translateHomeKey]
+            },
             {enabled: this.options.touch || this.options.mouse, class: CoordinateUtils},
             {enabled: this.options.keyboard, class: KeyboardEvents},
             {enabled: this.options.mouse, class: MouseEvents},
-            {enabled: this.options.gamepad, class: Gamepad, params: [this.gamepadManager, this.options.i18n]},
-            {enabled: this.options.camera, class: Camera, params: [this.options.i18n]},
+            {enabled: this.options.gamepad, class: Gamepad, params: [this.options.i18n], dependencies:[GamepadManager]},
```
pgivel, line 125: I see `GamepadManager` is not in this plugin map, nor the default one, except here as a dependency. Therefore it is loaded solely by the class name here. However the `MediaManager` is listed both directly as a dependency and as an item in the map.

I don't understand this inconsistency.

pgivel (changes requested):
From a a pure dev I like the mecanism (even if I have a couple things I don't understand in the implem)

However it doesn't seem like the simplest way to do it. Can't we simply move all dependencies in the default plugin list inside `DeviceRendererFactory`, and move all widgets to the custom plugin list inside `DeviceRender`, and order them as we like?

src/DeviceRendererFactory.js
jparez: it's to avoid to break actual implementation, cause actually addCustomPlugins load plugins

src/DeviceRendererFactory.js
jparez: Dep is the import of a class definition. The name attribute is the name of the class

src/DeviceRendererFactory.js
pgivel: Ok. I don't actually mind if we break the current implementation, since I'll be working on the Shadow player very soon and could update it next week. And in the meantime it wouldn't break stuff since the player version is pinned.

src/DeviceRenderer.js
jparez: Hum, not sure to correctly understand this part. I don't think this line `{enabled: this.options.microphone, class: MediaManager},` is require since microphone is only activate with camera, so the options.camera is enough.

src/DeviceRendererFactory.js
jparez: This way a widget couldn't be instanciate twice. We using a class structure but we bind instantiate object to the main "instance" varibale and so the widgets are use as singleton.

If we push plugin.class into dependenciesLoaded we couldn't instanciate a widget twice. Not sure that is a problem, but could be a source of bug :/

src/DeviceRendererFactory.js
jparez: As discussed, i prefere that addCustomPlugins stay a function. It would be cleaner if we need to put logic before return the array. Plus it could be use as a hook, but it's more a trick

pgivel (approved):

PR #80, opened 2024-06-12
[PLAYER-19] pre-keybinding refacto keyboard enable/disable and overlay

jparez:
## Description

PR before adding keymapping plugin. The goal is to have a cleaner way to manage the keyboard status.

- Overlay have been refacto to be manage by store
- Keyboard key press enable / disable have been refacto to be manage by store
- some eslint on a separate commit

## Type of change

- [x] Refacto

## Checklist

### DIFF
--- a/src/plugins/Battery.js
+++ b/src/plugins/Battery.js
@@ -110,8 +110,6 @@ module.exports = class Battery extends OverlayPlugin {
         this.widget.appendChild(close);
         this.widget.appendChild(this.container);
 
-        // Render into document
-        this.overlays.push(this.widget);
         this.instance.root.appendChild(this.widget);
     }
 
@@ -190,19 +188,6 @@ module.exports = class Battery extends OverlayPlugin {
     /**
      * Display or hide the widget.
      */
-    toggleWidget() {
-        // Notify other callers
-        if (this.widget.classList.contains('gm-hidden')) {
-            this.instance.emit('close-overlays');
-            this.instance.emit('keyboard-disable');
-        } else {
-            this.instance.emit('keyboard-enable');
-        }
-
-        // Toggle display
-        this.widget.classList.toggle('gm-hidden');
-        this.toolbarBtnImage.classList.toggle('gm-active');
-    }
 
     /**
      * Toggle Instance charging status between 'charging' and 'discharging';
--- a/src/plugins/util/OverlayPlugin.js
+++ b/src/plugins/util/OverlayPlugin.js
@@ -1,9 +1,9 @@
 'use strict';
+const {generateUID} = require('../../utils/helpers');
 
 /**
  * OverlayPlugin
- * Extendable utility to auto-bind and handle genymotion 'closeOverlays' event.
- * Expects implementer to provide list of overlays
+ * Parent for widget (plugin)
  */
 module.exports = class OverlayPlugin {
     /**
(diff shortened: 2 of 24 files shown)


src/plugins/util/OverlayPlugin.js
```diff
@@ -91,4 +93,12 @@ module.exports = class OverlayPlugin {
             this.restoreState();
         }
     }
+
+    toggleWidget(widget,toolbarBtnImage) {
+        this.instance.store.dispatch({type: 'OVERLAY_OPEN', payload: widget.classList.contains('gm-hidden')});
```
pgivel, line 98: maybe it would be interesting to say in the payload which widget was opened?

src/plugins/util/OverlayPlugin.js
```diff
@@ -91,4 +93,12 @@ module.exports = class OverlayPlugin {
...
+
+        // Toggle display
+        widget.classList.toggle('gm-hidden');
+        toolbarBtnImage.classList.toggle('gm-active');
```
pgivel, line 102: I'm not sure I understand why we pass the `toolbarBtnImage` as an arg instead of using `this.toolbarBtnImage` ?

src/plugins/util/OverlayPlugin.js
```diff
@@ -91,4 +93,12 @@ module.exports = class OverlayPlugin {
             this.restoreState();
         }
     }
+
+    toggleWidget(widget,toolbarBtnImage) {
+        this.instance.store.dispatch({type: 'OVERLAY_OPEN', payload: widget.classList.contains('gm-hidden')});
+
+        // Toggle display
+        widget.classList.toggle('gm-hidden');
```
pgivel, line 101: maybe instead of passing the widget as an arg we could iterate over `this.overlays`, like it's done in the `closeOverlays` function? (also maybe this new function and `closeOverlays` should be refactored and merged)

src/plugins/Battery.js
```diff
@@ -191,17 +191,7 @@ module.exports = class Battery extends OverlayPlugin {
...
-            this.instance.emit('close-overlays');
-            this.instance.emit('keyboard-disable');
-        } else {
-            this.instance.emit('keyboard-enable');
-        }
-
-        // Toggle display
-        this.widget.classList.toggle('gm-hidden');
-        this.toolbarBtnImage.classList.toggle('gm-active');
+        super.toggleWidget(this.widget, this.toolbarBtnImage);
```
pgivel, line 194: I'm not sure I understand why you changed this only in this BatteryPlugin ?

pgivel (changes requested):
I have a few small questions

src/plugins/util/OverlayPlugin.js
```diff
@@ -11,32 +12,45 @@ module.exports = class OverlayPlugin {
      * @param {Object} instance device renderer instance
      */
     constructor(instance) {
-        this.overlays = [];
-        this.savedState = null;
+        this.widget = null;
+        this.id = generateUID();
```
pgivel, line 16: I guess we could use a manual string id in the children classes, that way the store state would be more readable, but the way you've done it is nice too : we don't have to do anything when extending the OverlayPlugin, everything is already done here

src/plugins/util/OverlayPlugin.js
```diff
@@ -11,32 +12,45 @@ module.exports = class OverlayPlugin {
...
         this.instance = instance;
+        this.overlays = [];
 
         // Listen for close trigger
-        this.instance.registerEventCallback('close-overlays', this.closeOverlays.bind(this));
+        this.instance.store.subscribe(({overlay}) => {
+            if (overlay.widgetOpened.includes(this.id)) {
+                this.openOverlays();
+            } else {
+                this.closeOverlays();
```
pgivel, line 27: I think we can rename them to open/closeOverlay (without the 's' )

src/plugins/util/OverlayPlugin.js
```diff
@@ -11,32 +12,45 @@ module.exports = class OverlayPlugin {
...
      */
     constructor(instance) {
-        this.overlays = [];
-        this.savedState = null;
+        this.widget = null;
+        this.id = generateUID();
         this.toolbarBtnImage = null;
+        this.savedState = null;
         this.instance = instance;
+        this.overlays = [];
```
pgivel, line 20: 🔥 once every widget behaves the same way as the battery widget

src/store/index.js
pgivel, line 1: perfect,  and the comments are great

pgivel (approved):
I love it <3

pgivel (approved):
that's the kind of refacto I love <3 
image

PR #90, opened 2024-08-06
[PLAYER-35] change add event listener in apimanager

jparez:
## Description

This PR have several fix

- replace exposed API addEventlistener by the dedicated registerEventCallback function
- added closure on this.video for onConnectionClosed function? This way if we connect the VD more than 1 time this.video isn't undefined anymore
- remove gamepad unplug tracked event

### DIFF
--- a/src/DeviceRenderer.js
+++ b/src/DeviceRenderer.js
@@ -241,9 +241,13 @@ module.exports = class DeviceRenderer {
      * This will poll the WebRTC sequence until the WebSocket stops throwing the onclose event (unexpected quit).
      */
     onConnectionClosed() {
+        // closure to expose the right video and store context to onclose event
+        const video = this.video;
+        const store = this.store;
+
         this.webRTCWebsocket.onclose = (event) => {
-            this.store.dispatch({type: 'WEBRTC_CONNECTION_READY', payload: false});
-            this.video.style.background = this.videoBackupStyleBackground;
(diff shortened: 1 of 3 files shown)


src/DeviceRenderer.js
```diff
@@ -241,9 +241,12 @@ module.exports = class DeviceRenderer {
...
     onConnectionClosed() {
+        // closure to expose the right video context to onclose event
+        const video = this.video;
+
         this.webRTCWebsocket.onclose = (event) => {
             this.store.dispatch({type: 'WEBRTC_CONNECTION_READY', payload: false});
-            this.video.style.background = this.videoBackupStyleBackground;
+            video.style.background = this.videoBackupStyleBackground;
             this.initialized = false;
             log.debug('Error! Maybe your VM is not available yet? (' + event.code + ') ' + event.reason);
```
pgivel, line 251: I'm not sure I understand this: shouldn't the store also be moved outside the call in order for this to work? How about the `initialized` property?

I guess if it works it's fine, but this was mostly for my personal education :D

pgivel (approved):

src/DeviceRenderer.js
jparez: It's not mandatory cause store and initialized are always on object,(this.video is a HTML element) but you right, put all this variable in a closure is cleaner and could avoid futur issue

PR #91, opened 2024-08-06
[PLAYER-31] fix wrong calculation border with css scale rule

jparez:
## Description

When a css rules (i.e. scale) is set on the player's HTML element the calculation of borders is wrong. This is caused by the offset's browser api which doesn't take css rules in this calculation

- change way of borders are calculated
- fix bug on keymapping debug grids for portrait


pgivel (approved):
Nice !

another solution could have been to simply always remove the top border from the height and the left border from the width I guess (times two of course)

PR #95, opened 2024-08-08
[PLAYER-33] adding an example folder with a wrapper for the player

jparez:
## Description

Adding an "example" folder to the ./dist directory with files demonstrating the use of the player.

image


gulpfile.js
```diff
@@ -148,6 +174,7 @@ gulp.task('app-js', async function () {
     const bundler = await getBundler();
     return merge2(bundler.bundle().pipe(source(PATHS.SRC.APP)), {end: true})
         .pipe(gulpif(util.env.debug, using()))
+        .pipe(gulpif(util.env.production, replace("log.setDefaultLevel('debug');", "log.setDefaultLevel('error');")))
```
pgivel, line 177: This may break stuff a little bit for SiB dev & debug, but I guess it's OK, we'll have to handle it on our side. I may need to talk with you at a later time if I get stuck on it.

pgivel (approved):

PR #100, opened 2024-09-10
replace way of check if plugin charged

jparez:
## Description

When player is used in third party app class.name could be empty in case of uglify process on node modules

## Type of change


pgivel (approved):
this seems like a good enough solution :)

PR #104, opened 2024-09-19
[PLAYER-39] adding player version to js and css minified files

jparez:
## Description

Adding version to minified files

## Type of change

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -31,6 +31,7 @@ const uglify = require('gulp-uglify-es').default;
 const using = require('gulp-using');
 const util = require('gulp-util');
 const replace = require('gulp-replace');
+const header = require('gulp-header');
 
 const PATHS = {
     SRC: {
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -97,15 +98,6 @@ gulp.task('app-partials', function () {
...
 
-    gulp.src(['./example/geny-window.html'])
-        .pipe(
-            replace(
-                /https:\/\/cdn\.jsdelivr\.net\/npm\/@genymotion\/device-web-player@[\d\.]+/g,
-                `https://cdn.jsdelivr.net/npm/@genymotion/device-web-player@${version}`,
-            ),
-        )
-        .pipe(gulp.dest('./example/'));
-
```
pgivel, line 108: I'm still not sure why you want to get rid of this replace step?

pgivel (commented):
There's one change I don't understand, LGTM otherwise

gulpfile.js
jparez: in line below we parse all example folder, so this code is executed twice

pgivel (approved):

PR #108, opened 2024-09-23
Closing & reopening the webrtc connection loses the mouse & keyboard inputs

skornous:
## Description

Since the rework of v4 a bug was present where if you closed & reopened the WebRTC WebSocket you'd lose the inputs from both mouse & keyboard.

After some tests it appears that unregistering the mouse & keyboard events in the store on WebSocket disconnection & then re-registering them when creating the peer connection (as for the touch & gamepad events) fixes this issue.

This PR is the aforementioned fix.

Sorry but as it's quite a technical issue, it's hard to provide a simple screen-record so you'll have to trust me on this 😬 

Fixes #(no related issue)

## Type of change

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

-   [x] I've read & comply with the [contributing guidelines](https://github.com/Genymobile/genymotion-device-web-player/blob/main/CONTRIBUTING.md)
-   [x] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers.
-   [x] I have made corresponding changes to the documentation (README.md).
-   [x] I've checked my modifications for any breaking changes.

PR #109, opened 2024-09-24
wip

pgivel:
## Description

related to #108


src/DeviceRenderer.js
```diff
@@ -478,6 +485,17 @@ module.exports = class DeviceRenderer {
...
+            // in case this is a re-start, we want to add the callbacks again depending on the state
+            if (this.store.get('KEYBOARD_EVENTS_ENABLED')) {
+                this.keyboardEvents.addKeyboardCallbacks();
+            }
+            if (this.store.get('MOUSE_EVENTS_ENABLED')) {
+                this.mouseEvents.addMouseCallbacks();
+            }
+            if (!this.store.get('KEYBOARD_EVENTS_ENABLED') && !this.store.get('MOUSE_EVENTS_ENABLED')) {
+                this.keyboardMapping.addKeyboardCallbacks();
+            }
```
pgivel, line 497: These getters don't actually work for now, this is for discussion purposes. A working diff would look like this, but I don't think these state properties are supposed to be public-facing?
```suggestion
            // in case this is a re-start, we want to add the callbacks again depending on the state
            if (this.store.state.isKeyboardEventsEnabled) {
                this.keyboardEvents.addKeyboardCallbacks();
            }
            if (this.store.state.isMouseEventsEnabled) {
                this.mouseEvents.addMouseCallbacks();
            }
            if (!this.store.state.isKeyboardEventsEnabled && !this.store.state.isMouseEventsEnabled) {
                this.keyboardMapping.addKeyboardCallbacks();
            }
```

jparez: this.store.state is the right way to get values from the store

PR #112, opened 2024-09-26
Fix the example: remove api token error when we fetch recipes

caarmen:
## Description

This fixes a minor bug where we had an error message in the recipe selection list.

Steps to reproduce:
* open the example page
* copy your api token
* expected behavior: the recipes list is populated
* actual behavior: the recipes list is populated BUT there's still an error about a missing api token

Before:
image

With the fix:
image

## Type of change

-   [ ] New feature (non-breaking change which adds functionality)

## Checklist

-   [x] I've read & comply with the [contributing guidelines](https://github.com/Genymobile/genymotion-device-web-player/blob/main/CONTRIBUTING.md)
    - Not really, it's "just" an example 🤞🏻 😅 
    - I don't think anything is needed.
    - Not really, it's "just" an example 🤞🏻 😅

### DIFF
--- a/example/geny-window.js
+++ b/example/geny-window.js
@@ -155,6 +155,8 @@ const fetchRecipes = async () => {
         }
 
         const selectElement = document.querySelector('#listRecipes');
+        // Empty the select before adding new options.
+        selectElement.innerHTML = '';
         const placeholderOption = document.createElement('option');
         placeholderOption.textContent = 'Select a recipe';
         placeholderOption.value = '';


example/geny-window.js
```diff
@@ -155,6 +155,8 @@ const fetchRecipes = async () => {
         }
 
         const selectElement = document.querySelector('#listRecipes');
+        // Remove our "warning" select option about an invalid/missing api token.
+        selectElement.remove(0);
         const placeholderOption = document.createElement('option');
```
jparez, line 160: well play, but the bug is worth that what you detect. Every time the input apitoken loose focus, the recipes list is added to the previous.

This will fix all issues
```
const selectElement = document.querySelector('#listRecipes');
     // Flush select before adding new options
      selectElement.innerHTML = '';
 const placeholderOption = document.createElement('option');
```

jparez (changes requested):

example/geny-window.js
caarmen: > well play, but the bug is worth that what you detect. Every time the input apitoken loose focus, the recipes list is added to the previous.

Ah, nice catch! 😅 

I've adjusted the fix with your suggestion. Thanks!

jparez (approved):
LGTM, thank for this PR !

pgivel (approved):

PR #118, opened 2025-01-06
[PLAYER-42] adding theme capability to player

jparez:
## Description

adding theme capability to player

image
image
image
image
image

## Type of change

## Checklist

-   [ ] I've read & comply with the [contributing guidelines](https://github.com/Genymobile/genymotion-device-web-player/blob/main/CONTRIBUTING.md)


README.md
```diff
@@ -118,6 +118,51 @@ or check the [PaaS documentation](https://docs.genymotion.com/paas/01_Requiremen
 </script>
 ```
 
+## Style and CSS
+
+The player use css variables to stylized some parts of the app, but the player is still totally customisable throught overload of the css class.
```
pgivel, line 123:
```suggestion
The player use css variables to stylize some parts of the app, but the player is still totally customisable through overloading the css classes.
```

README.md
```diff
@@ -118,6 +118,51 @@ or check the [PaaS documentation](https://docs.genymotion.com/paas/01_Requiremen
...
+
+#### CSS variables available are theirs default presets:
```
pgivel, line 125:
```suggestion
#### CSS variables available with their default presets:
```

src/scss/base/_genymotion.scss
```diff
@@ -49,8 +53,8 @@
             position: absolute;
             z-index: 1;
             cursor: pointer;
-            background-color: $black;
-            color: $white;
+            background-color: #000;
```
pgivel, line 56: shouldn't this be defined through theme too?

src/scss/components/_battery.scss
```diff
@@ -49,7 +50,7 @@
                     bottom: 3px;
                     width: 80%;
                     height: 0;
-                    background-color: $white;
+                    background-color: #fff;
```
pgivel, line 53: I feel this should be defined by the theme too, but I guess that's what the "TODO" above means? Maybe you could expand what exactly is to be done, and possibly refer to a jira ticket id?

src/scss/components/_clipboard.scss
```diff
@@ -6,8 +6,10 @@
         &-clipboard-input {
             min-height: 300px;
             resize: none;
-            color: $black !important;
-            background-color: $white !important;
+
+            /* Todo make better */
```
pgivel, line 10: lol this is better than simply "todo", but still not super explicit

src/scss/components/_turn.scss
```diff
@@ -7,14 +7,15 @@
         
         position: absolute;
         left: -305px;
-        color: $white;
+        //TODO maquette ceci, pour afficher lancer displayDefaultTurnWarning() et desactiver hideDefaultTurnWarning
```
pgivel, line 10: je ne comprends pas ce commentaire

pgivel (changes requested):
Great feature <3

README.md
caarmen: Also: "The player uses" (instead of "The player use").

"style" might also be more straightforward than "stylize".

example/geny-window.css
```diff
@@ -18,6 +18,10 @@
     margin-left: 5px;
 }
 
+.mr-1 {
```
caarmen, line 21: I think vuetify and tailwind use 4px for `mr-1`. Not sure if this is a more global standard. Is it ok to have a slightly different value?

Actually, this is just the example css, should be ok :)

src/scss/components/_fingerprints.scss
```diff
@@ -33,7 +33,7 @@
                 &.active {
                     pointer-events: all;
                     background-color: #152616;
-                    color: $succes-color;
+                    color: var(--succes-color);
```
caarmen, line 36: Looks like the css variable has the correct spelling where it's declared (success).
```suggestion
                    color: var(--success-color);
```

src/scss/base/_variables.scss
```diff
@@ -1,33 +1,48 @@
+/**
+* CSS variable
```
caarmen, line 2:
```suggestion
* CSS variables
```

src/scss/components/_fingerprints.scss
```diff
@@ -71,10 +71,19 @@
...
                     width: 80%;
                     cursor: pointer;
-                    &:hover {
-                        background: $btn-hover-background-color;
+                    position: relative;
+                    z-index: 0;
+                    &:hover::before {
+                        content: '';
+                        position: absolute;
+                        top: 0;
```
caarmen, line 79: Could you please explain a bit how the positioning changes for the fingerprint dialog button relate to the theme capability? (For my understanding)

src/scss/components/_gps.scss
```diff
@@ -31,7 +31,7 @@
             left: 26px;
             right: 26px;
             top: 26px;
-            background-color: $white;
+            background-color: #fff;
```
caarmen, line 34: Perhaps another color to put into a variable?

caarmen (commented):
Cool new feature!

src/scss/base/_genymotion.scss
jparez: There is much more to do here, i've created a jira for this task https://genymobile.atlassian.net/jira/software/projects/PLAYER/boards/193?selectedIssue=PLAYER-52

src/scss/components/_battery.scss
jparez: this is a part of next story (rework battery modal)

src/scss/components/_clipboard.scss
jparez: This is a part of next story (rework clipboard modal). BTW i can remove the comment

src/scss/components/_fingerprints.scss
jparez:
```
As of 2025, you typically no longer need to prefix mask-image with -webkit- for most modern browsers, as support has improved significantly. However, there are some considerations:
Current Browser Support

    Modern Browsers (Chrome, Edge, Firefox, Safari):
        Chrome/Edge: mask-image is fully supported without the -webkit- prefix.
        Safari: As of the latest versions, mask-image support still requires the -webkit- prefix for compatibility.
        Firefox: Fully supports mask-image without a prefix.

    Legacy Browsers:
        If you need to support older versions of Safari or other WebKit-based browsers, you should include the -webkit- prefix.

Recommendation

For maximum compatibility, especially if your audience includes Safari users or older WebKit-based browsers, it’s good practice to include both prefixed and unprefixed versions:
```

caarmen: As discussed offline, my understanding is that, for the fingerprint button in hover state: we're replacing a background attribute with a pseudo-element. This element is positioned to match the position/size of the button, and the hover background is now controlled on this pseudo-element instead of on the button itself.

It's related to the changes in the svg images, and the way the background is set on the svg images. I don't understand it completely but that's the high level I grasped.

caarmen (approved):
Nice to have this theming capability.

Maybe consider putting a more descriptive comment for the TODO, but not blocking.

PR #120, opened 2025-01-13
[PLAYER-49] refacto disk io

jparez:
## Description

- adding readonly and fix bug on regexFilter. Refacto swicth setState
- dropdown and chipTag components
-  Rework ioThrottling plugin UI / UX in accordance to figma

image


src/assets/images/ic_check.svg
pgivel, line 1: Is it normal that the checkmark isn't centered in this svg?

pgivel (approved):

PR #124, opened 2025-02-10
[Player-54] refacto modal position

jparez:
## Description

- Add an option to display the toolbar on the left or right side of the screen
- Calculate the position of the modal when opening for the first time
- Make the widget modal draggable inside the video wrapper
- Add a magnetic feature when the modal is dragged near its starting position
- Small refactor and fixes (icons, color theme, etc.)

Demo

https://github.com/user-attachments/assets/f9079d16-6ffe-438b-9af5-c5936b1cfcc0

## Type of change


README.md
```diff
@@ -326,6 +326,13 @@ A device renderer instance can be configured using the `options` argument (objec
 -   **Details:**
     Adds a mobile-style frame around the video to mimic the appearance of a smartphone screen.
 
+### `toolbarPosition`
+
+-   **Type:** `String`
+-   **Default:** `right`
+-   **Details:**
```
caarmen, line 333: I guess the other possible values are top, left, bottom? Might help to mention it just in case.

src/DeviceRendererFactory.js
```diff
@@ -86,6 +87,7 @@ module.exports = class DeviceRendererFactory {
      * @param  {string}             webRTCUrl                      WebRTC URL of the instance.
      * @param  {Object}             options                        Various configuration options.
      * @param  {boolean}            options.showPhoneBorder        Show phone border. Default: false.
+     * @param  {string}             options.toolbarPosition        Toolbar position. Default: 'right'.
```
caarmen, line 90: perhaps document other possible values.

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
-     *
-     * @param {Object} [options={}] - Options for the modal.
-     * @param {string|null} [options.title=null] - The title of the modal. If null, no title will be set.
-     * @param {string} [options.classes=''] - Additional classes to add to the modal container.
-     * @returns {Object} An object containing the modal and container elements.
-     * @returns {HTMLElement} return.modal - The modal element.
-     * @returns {HTMLElement} return.container - The container element inside the modal.
+     * Creates a modal template with configurable options
+     * @param {Object} [options={}] - Modal configuration options
+     * @returns {Object} Modal elements container
```
caarmen, line 90: Is this `@returns` documentation correct? I see this:
```javascript
return {modal: modalElement, container: modalBody};
```

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
+        modalElement.appendChild(modalHeader);
+        modalElement.appendChild(modalBody);
+
+        this.initializeModalBehavior(modalElement, modalHeader);
+        return {modal: modalElement, container: modalBody};
+    }
+
+    /**
+     * Creates modal header element with title and close button
+     * @param {?string} title - Header title content
```
caarmen, line 108: What's preferable ? `?string` or `string|null`?

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
+     * @returns {HTMLElement} Configured header element
+     */
+    createModalHeader(title) {
+        const header = document.createElement('div');
```
caarmen, line 112: Why not use a semantic element like [`header`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header)? Is it out of scope for this PR?

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
+
+        return header;
+    }
+
+    /**
+     * Creates modal close button element
+     * @returns {HTMLElement} Configured close button
+     */
+    createCloseButton() {
+        const button = document.createElement('div');
```
caarmen, line 131: Why not use a `<button>` element?

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
+
+    /**
+     * Creates base modal element structure
+     * @param {string} classes - Additional CSS classes
+     * @param {?number} width - Modal width in pixels
+     * @param {?number} height - Modal height in pixels
+     * @returns {HTMLElement} Configured modal element
+     */
+    createModalElement(classes, width, height) {
+        const modal = document.createElement('div');
```
caarmen, line 155: [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) wouldn't be relevant?

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
-        divHeader.className = 'gm-modal-header';
-        const divTitle = document.createElement('div');
-        divTitle.className = 'gm-modal-title';
-        divTitle.innerHTML = title;
-        divHeader.appendChild(divTitle);
+        return modal;
+    }
+
+    /**
+     * Initializes drag-and-drop and positioning for the modal
```
caarmen, line 169: Is this "drag and drop"? Or just dragging (positioning/moving)?

Is there an action done upon dropping?

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
+     * Enables drag-and-drop functionality for the modal
+     * @param {HTMLElement} modal - Modal element to make draggable
+     * @param {HTMLElement} dragHandle - Element that will trigger dragging
+     */
+    enableDragBehavior(modal, dragHandle) {
+        let isDragging = false;
+        dragHandle.style.cursor = 'move';
 
-        divModal.appendChild(divBody);
+        this.instance.addListener(dragHandle, 'mousedown', (event) => {
```
caarmen, line 194: Do we ever have to remove this mousedown listener? No chance of a memory leak?

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,114 +2,353 @@
...
+                removeMouseMoveListener();
+                removeMouseUpListener();
+            });
+        });
+    }
+
+    /**
+     * Handles drag start initialization
+     * @param {HTMLElement} modal - Modal element
+     * @param {MouseEvent} event - Mouse event
```
caarmen, line 244: Is this event jsdoc param right?

src/scss/base/_genymotion.scss
```diff
@@ -165,7 +165,7 @@
             max-height: 100%;
 
             margin: auto;
-            // Todo calc via JS
+            // calc throught JS
```
caarmen, line 168: through?

src/plugins/util/OverlayPlugin.js
jparez: Like you said, it's a semantic element used for SEO purposes. The <header> tag can help improve code understanding for a main header, but I'm not sure it's useful for the readability of a modal. Personally, I prefer using <div> elements.

src/plugins/util/OverlayPlugin.js
jparez: Because browsers can style button elements, but in this case, it's just an icon.

src/plugins/util/OverlayPlugin.js
caarmen: It's for SEO but also accessibility (screen readers for the visually impaired).

README.md
jparez: Only 'left' and 'right' are supported. For 'top' and 'bottom', additional code is required to calculate the position. This could be a future feature.

src/plugins/util/OverlayPlugin.js
jparez: string|null is clearer

src/plugins/util/OverlayPlugin.js
jparez: Right, but the app isn't ARIA-compliant, so this alone isn't really relevant.

src/plugins/util/OverlayPlugin.js
jparez: It could be, since the <dialog> element comes with a built-in API for opening and closing. However, given the legacy code, I'm not sure it's worth the effort to refactor. For a new project, though, it would definitely be a better choice.

src/plugins/util/OverlayPlugin.js
jparez: It's just dragging, but in my confused mind, 'drag' always comes with 'drop'! >_<

src/plugins/util/OverlayPlugin.js
jparez: It's always a good idea to remove listener to avoid memory leak.

In DeviceRenderer.js we call `this.removeAllListeners();` in the `destroy()` function. So if client call the disconnect function exposed by API we are good

```
this.registerFunction({
            name: 'disconnect',
            category: 'VM_communication',
            fn: () => {
                this.instance.destroy();
            },
            description: 'Disconnect from the current instance, ending the WebSocket communication.',
});
```
By the way description could be improved to indicate that this fn clean all listener

src/plugins/util/OverlayPlugin.js
jparez: The code has evolved, but the documentation hasn't

src/plugins/util/OverlayPlugin.js
```diff
@@ -2,108 +2,358 @@
...
-    closeOverlay() {
-        if (this.widget && !this.widget.classList.contains('gm-hidden')) {
-            this.widget.classList.add('gm-hidden');
-            if (this.widget.onclose) {
-                this.widget.onclose();
-            }
+    calculateModalPosition(triggerElement, toolbarPosition) {
+        const triggerRect = triggerElement.getBoundingClientRect();
+        const modalWidth = this.widget.offsetWidth || 200;
+        const modalHeight = this.widget.offsetHeight || 100;
```
pgivel, line 277: Maybe these values could be extracted to a CONST at the begining of this file. Similar comment for the +/- 10 values in the next lines of this function

pgivel (changes requested):

src/plugins/IOThrottling.js
```diff
@@ -191,7 +186,7 @@ module.exports = class IOThrottling extends OverlayPlugin {
         this.container.appendChild(clearCacheDiv);
 
         // Render into document
-        this.instance.root.appendChild(this.widget);
+        this.instance.root.appendChild(modal);
```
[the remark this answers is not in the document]
jparez: for IoThrottling is a mistake :$

src/plugins/util/OverlayPlugin.js
caarmen: Ok, thanks!

src/plugins/util/OverlayPlugin.js
caarmen: The context of that sentence: the doc is saying to use button or input elements, instead of the aria role "button" with a div. 

I think we're not using the aria role button, but that just means using a button element would be even more helpful, than no indication at all that the close icon serves functionally as a button.

src/plugins/KeyboardEvents.js
```diff
@@ -225,11 +223,20 @@ module.exports = class KeyboardEvents {
...
+                            this.instance.video.focus();
+                        },
+                        removeListener: null,
+                    },
                 ];
             }
             this.instance.root.focus();
             this.keyboardCallbacks.forEach((item, index, array) => {
-                array[index].removeListener = this.instance.addListener(this.instance.root, item.event, item.handler);
+                array[index].removeListener = this.instance.addListener(this.instance.video, item.event, item.handler);
```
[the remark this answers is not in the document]
pgivel: @jparez After discussing a bit with Dimitri, there is no issue with breaking SiB compatibility, we just have to list the breaking changes in the release changelog. (I think changing the html element that listens for the events is probably a breaking change)

pgivel (approved):

PR #128, opened 2025-03-18
[PLAYER-46] refacto network widget

jparez:
## Description
Refacto network plugin

- Plugin is now disable for android below version 9 (based on data.message.mobileThrottling from capabilites)

image


src/plugins/CoordinateUtils.js
```diff
@@ -113,7 +113,12 @@ module.exports = class CoordinateUtils {
     getXCoordinate(event) {
         this.instance.root.focus(); // not on the Y function because we need to focus element only once
         this.setVideoElement();
-        const x = this.getRawXCoordinateFromEvent(event) - this.getLeftBorder();
+        const style = window.getComputedStyle(this.video);
+
+        // Retrieve the border size (converted to a number) if it exists
+        const borderLeft = parseFloat(style.borderLeftWidth) || 0;
+        const x = this.getRawXCoordinateFromEvent(event) - this.getLeftBorder() + borderLeft;
```
pgivel, line 120: What's the reason for removing and adding the left border? This is not super explicit, a comment could help (same for yCoordinate)

src/plugins/NetworkOLD.js
```diff
@@ -0,0 +1,601 @@
+'use strict';
+
+const log = require('loglevel');
+log.setDefaultLevel('debug');
+
+const OverlayPlugin = require('./util/OverlayPlugin');
+
+const PROFILES = require('./util/network-profiles');
```
pgivel, line 8: This file (`plugins/util/network-profiles`) has been deleted in the same PR

pgivel (changes requested):

src/assets/images/ic_network-type-2G-edge.svg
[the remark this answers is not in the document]
jparez: For 2 things : 
- there is an unlimited icon, so to keep coherence
- to have same typo that design

src/plugins/NetworkOLD.js
jparez: oups

src/plugins/CoordinateUtils.js
jparez: Adding the border causes a position break, so here we check if there is a border on the video to include it in the calculation.

image


I will make a comment on code to explain that

pgivel (approved):

PR #135, opened 2025-04-28
[PLAYER-70] Refacto screencast widget

jparez:
## Description
screen record and screen shot are now two separate button in the toolbar instead of a single plugins with this 2 buttons

Before:
- screencast widget was using a single button for both screenshot and screenrecord
- Time of screenrecord was displayed near toolbar in a tooltip

After:
- screencast widget deleted
- options to activate screenshot and screenrecord are still 'capture' (option activated or desactivated both plugins)
- screen record and screen shot are now two separate button in the toolbar
- time of screenrecord is now displayed in the toolbar (below the record button)
- when screenrecord is active, the background of the player is changed

Screenrecord and screenshot buttons can be included in the toolbar or in the floating toolbar Adding tests for screenshot and screenrecord buttons

Please include a summary of the change and which issue is fixed if any. Please also include relevant motivation and context (a screenshot is appreciated).

## Type of change

-   [ ] New feature (non-breaking change which adds functionality)

## Checklist

-   [ ] I've read & comply with the [contributing guidelines](https://github.com/Genymobile/genymotion-device-web-player/blob/main/CONTRIBUTING.md)


src/plugins/Screenrecord.js
```diff
@@ -0,0 +1,233 @@
...
+            'video/ogg': '.ogv',
+            'video/webm': '.webm',
+            'video/3gpp': '.3gp',
+            'video/3gpp2': '.3g2',
+            'video/mp4': '.mp4',
+            'video/x-matroska': '.mkv',
+            'video/x-flv': '.f4v',
+        };
+
+        let extension = types[mime.toLowerCase()];
```
caarmen, line 92: In which scenario could we have a `mime` argument which is different from the values we look for in `findBestMimeType`?

src/plugins/Screenrecord.js
```diff
@@ -0,0 +1,233 @@
...
+        const endTime = new Date();
+
+        // compute seconds
+        let timeDiff = endTime - this.startTime;
+        timeDiff /= 1000;
+        const seconds = Math.round(timeDiff % 60);
+
+        // compute minutes
+        timeDiff = Math.floor(timeDiff / 60);
+        const minutes = Math.round(timeDiff % 60);
```
caarmen, line 141:
```suggestion
        const minutes = Math.floor(timeDiff / 60);
```

Since we're comparing `MAX_SCREENRECORD_LENGTH_IN_MINUTES` to the duration in minutes, the modulo would provide incorrect results if  `MAX_SCREENRECORD_LENGTH_IN_MINUTES` were bigger than one hour (it's not the case today, so no bug for now).

Example, let's say the elapsed time is 63 minutes and 27 seconds:
endTime - startTime = 3807000.

```javascript
        let timeDiff = endTime - this.startTime; // 3807000
        timeDiff /= 1000; // 3807
        const seconds = Math.round(timeDiff % 60); //27

        // compute minutes
        timeDiff = Math.floor(timeDiff / 60); // 63
        const minutes = Math.round(timeDiff % 60); // 3
 ```

We would compare `MAX_SCREENRECORD_LENGTH_IN_MINUTES` to 3 minutes, instead of to 63 minutes, without the suggested change.

Comment (caarmen) [reply src/plugins/Screenrecord.js:141]: Feel free to verify the example above :) Didn't test out the code, just used a calculator 😅

File: example/geny-window.js [hunk 3]
```diff
@@ -117,21 +117,31 @@ const initPlayer = (webrtcAddress) => {
     const options = {
         token: jwtToken || apiToken,
         showPhoneBorder: true,
+        toolbarPosition: 'right',
```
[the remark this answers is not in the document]
jparez: this is an option to display toolbar on the left or on ther right, this is unecessary because the position of the toolbar is on the right by default

src/plugins/Screenrecord.js
jparez: For example if the browser might not support the exact MIME type we requested or it add additional parameters to the MIME type

caarmen (approved):
📸 🎥 👍🏻

PR #149, opened 2025-08-19
[PLAYER-84] adding accesToken generation to readme

jparez:
## Description

 - adding accesToken generation to readme
 - Change VM top padding from 30px to 25px


README.md
```diff
@@ -128,6 +130,12 @@ or check the [PaaS documentation](https://docs.genymotion.com/paas/01_Requiremen
 </script>
 ```
 
+### 🔑 A word about the token
+
+#### Saas
+
+To connect a [SaaS](https://docs.genymotion.com/saas/) instance you need a valid token. This token must be generated by using your [endpoint API](https://developer.genymotion.com/saas/#tag/Instances-v1/operation/accessToken). When using this API, don't forget to include the x-api-token HTTP header with a valid [Bearer](https://developer.genymotion.com/saas/#tag/Users-v1/operation/login) or [APIToken](https://www.genymotion.com/blog/api-tokens-for-genymotion-saas/)
```
jmaneyrol69: It may not be very clear for people who are not very familiar. I thing you should make it clear that the generated token is _not_ the same token as the API token.
Maybe something like this:
```suggestion
To connect a [SaaS](https://docs.genymotion.com/saas/) instance you need to generate and use a specific token. This must be done by using our [endpoint API](https://developer.genymotion.com/saas/#tag/Instances-v1/operation/accessToken). When using this API, don't forget to include the x-api-token HTTP header with a valid [Bearer](https://developer.genymotion.com/saas/#tag/Users-v1/operation/login) or [API Token](https://www.genymotion.com/blog/api-tokens-for-genymotion-saas/)
```

jmaneyrol69 (changes requested):

README.md
jmaneyrol69: I also think you should add a code example here as well.

jmaneyrol69 (approved):
LGTM

PR #159, opened 2025-12-23
[Player-59] camera widget

jparez:
## Description

This branch introduces camera injection management and improves the local development workflow.

Front and back camera are implemented but back camera is hidden, waiting for system handle more than one stream

### Camera Functionality (feat(Camera))
- Camera Plugin: Added Camera.js to handle video stream injection.
- Media Manager: Updated MediaManager.js to support new sources (Dummy, local video) and front / back camera
- **User Interface:**
   - Added the GmDropdown component for source selection.
   - Integrated a fileUploader for manage streaming custom video files.
   - Added / updated associated SCSS styles (_camera.scss, _fileUploader.scss).
- Testing: Added unit tests (camera.test.js).

### Configuration & Dev (chore(vite))
- HMR Improvements: Adjusted vite.config.js with an html-transform plugin to optimize Hot Module Replacement. This prevents conflicts during script/style injection, ensuring better stability while developing the camera features.

image
image


vite.config.js
```diff
@@ -2,9 +2,67 @@ import { defineConfig } from 'vite';
...
     return {
-        // plugins: [viteSingleFile()],
+        plugins: [
+            basicSsl(),
+            {
+                name: 'html-transform',
+                transformIndexHtml(html) {
+                    if (mode === 'development') {
+                        let newHtml = html.replace(
+                            /<script src="\.\.\/dist\/js\/device-renderer\.min\.js" data-player-url><\/script>/,
```
caarmen, line 16: Would it be possible to move this code into another file, to keep the `vite.config.js` as lean and as declarative as possible?

src/plugins/util/fileUploader.js
```diff
@@ -278,17 +281,37 @@ const fileUploader = (() => {
...
         const checkFileBeforeUpload = (file) => {
             if (file) {
-                if (!accept || (accept && file.name.toLowerCase().endsWith(accept.toLowerCase()))) {
+                let isValid = false;
+                if (!accept) {
+                    isValid = true;
+                } else {
+                    const acceptedTypes = accept.split(',').map((t) => t.trim().toLowerCase());
+                    isValid = acceptedTypes.some((type) => {
+                        if (type.endsWith('/*')) {
```
caarmen, line 291: Could we have one example of a `type` value, in a comment, in each of these `if` branches?

src/plugins/util/fileUploader.js
```diff
@@ -34,6 +35,8 @@ const fileUploader = (() => {
         accept = null,
         classes = '',
         i18n = {},
+        mode = 'upload',
```
caarmen, line 39: I don't see `mode` in the jsdoc. Did I miss something?

I see that in `Camera.js`, we specify `mode: "select"`. Would be nice to document the two possible modes in the jsdoc.

vite.config.js
```diff
@@ -2,9 +2,67 @@ import { defineConfig } from 'vite';
 import { viteSingleFile } from 'vite-plugin-singlefile';
 import path from 'path';
 
+import basicSsl from '@vitejs/plugin-basic-ssl';
+
 export default defineConfig(({ mode }) => {
     return {
-        // plugins: [viteSingleFile()],
+        plugins: [
+            basicSsl(),
```
caarmen, line 9: Is this `basicSsl` related to the hot reload?

I see that an ssl dependency is added, but not in this commit which introduces the plugin.

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+                if (sourceWidth && sourceHeight) {
+                    ctx.drawImage(mediaElement, 0, 0, sourceWidth, sourceHeight);
+                }
+            };
+
+            let audioTrack = null;
+            if (this.instance.options.microphone && !isImage) {
+                try {
+                    if (!this.audioContext) {
+                        this.audioContext = new (window.AudioContext || window.webkitAudioContext)();
```
caarmen, line 533: Is this `AudioContext` ever [close](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/close)d?

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+
+                    const audioTracks = dest.stream.getAudioTracks();
+                    log.debug(`[Camera] Audio tracks captured: ${audioTracks.length}`);
+                    if (audioTracks.length > 0) {
+                        audioTrack = audioTracks[0];
+                        audioTrack.enabled = true;
+                    }
+
+                    this.audioNodes = this.audioNodes || [];
+                    this.audioNodes.push({source, dest, type});
```
caarmen, line 552: Copilot mentions that this audioNodes array might have a potential memory leak. Not completely sure to understand: maybe because we put source nodes and destination nodes in the array, and when we're done with the camera, we don't clean the array? We'll have dangling references to these source/destination nodes?

src/components/GmDropdown.js
```diff
@@ -156,7 +156,7 @@ export class GmDropdown extends HTMLElement {
         if (typeof item === 'string') {
             return item;
         }
-        return item.valueToDisplay ?? item.element?.textContent ?? item.value;
+        return item.label ?? item.valueToDisplay ?? item.element?.textContent ?? item.value;
```
caarmen: Why do we need to support a new attribute `label`?

Why can't `Camera.js` supply `valueToDisplay`? It seems to have the same results.

Could this component have some documentation in `set items(newItems)` explaining what are the accepted types of objects that can be passed for `newItems`?

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+
+        // 1. None
+        videoOptions.push({value: 'none', label: 'No media'});
+
+        // 2. Webcams (if any)
+        this.videoDevices.forEach((d) => {
+            if (d.label) {
+                videoOptions.push({
+                    value: d.deviceId,
+                    label: d.label,
```
caarmen: Why not supply `valueToDisplay: d.label` instead? (Same a few lines above).

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+        if (this.backDropdown) {
+            this.backDropdown.items = videoOptions;
+            if (!this.selectedVideoDevice) {
+                this.backDropdown.value = 'none';
+                this.onDeviceSelected('back', 'none');
+            }
+        }
+    }
+
+    onFileSelected(file, type) {
```
caarmen, line 397: Not sure where to put this:
If I select the same file multiple times, it only seems to work the first time.

caarmen: https://github.com/user-attachments/assets/9d808aea-f263-416b-be02-dd02c17fa4a5

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+
+        const requestText = document.createElement('div');
+        requestText.className = 'gm-permission-text';
+        requestText.innerHTML =
+            this.i18n.CAMERA_GRANT_ACCESS_TEXT || 'Grant camera and microphone access to use media injection.';
+
+        this.grantBtn = document.createElement('button');
+        this.grantBtn.className = 'gm-btn'; // Reusing common button class
+        this.grantBtn.innerHTML = this.i18n.CAMERA_GRANT_ACCESS_BTN || 'GRANT ACCESS';
+        this.grantBtn.onclick = () => this.requestPermission();
```
caarmen, line 110: Am I missing something? I tried to grant access to the camera, and seems that the access wasn't detected by the app.

https://github.com/user-attachments/assets/c67bf302-f58b-43ac-a3ef-63e6e771337e

caarmen: I added a log:
```diff
diff --git a/src/plugins/Camera.js b/src/plugins/Camera.js
index 13414af..85d441d 100644
--- a/src/plugins/Camera.js
+++ b/src/plugins/Camera.js
@@ -253,6 +253,7 @@ export default class Camera extends OverlayPlugin {
         try {
             const devices = await navigator.mediaDevices.enumerateDevices();
             this.videoDevices = devices.filter((d) => d.kind === 'videoinput');
+            console.log('Video devices found:', this.videoDevices);

```

I see these logs when I click on "allow while visiting the site" or "allow this time":

Video devices found: [InputDeviceInfo]0: InputDeviceInfodeviceId: ""groupId: ""kind: "videoinput"label: ""[[Prototype]]: InputDeviceInfolength: 
onDeviceSelected called: type=front, deviceId=none
onDeviceSelected called: type=back, deviceId=none

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+
+                    if (this.backDropdown) {
+                        this.backDropdown.value = firstDevice.deviceId;
+                    }
+                }
+            }
+        } catch (error) {
+            this.permissionRequestView.classList.add('hidden');
+
+            if (error.name === 'NotFoundError' || error.message.includes('The object can not be found')) {
```
caarmen: Will these value `NotFoundError` and  `The object can not be found` be consistent across browsers?

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+        if (this.frontDropdown) {
+            this.frontDropdown.items = videoOptions;
+            if (!this.selectedVideoDevice) {
+                this.frontDropdown.value = 'none';
+                this.onDeviceSelected('front', 'none');
+            } else {
+                // Try to preserve selection if it still exists or default to none
+                const exists = videoOptions.some((o) => o.value === this.selectedVideoDevice);
+                if (!exists && this.selectedVideoDevice !== 'file') {
+                    // 'file' might not be in options explicitly if we dynamic it, but here it is static
```
caarmen: Not sure what "if we dynamic" it means, could you clarify?

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+        const fileNameElement = type === 'front' ? this.frontFileName : this.backFileName;
+        if (fileNameElement) {
+            fileNameElement.innerText = file.name;
+        }
+
+        const url = URL.createObjectURL(file);
+        this.startFileStream(type, url, file);
+    }
+
+    async startFileStream(type, url, file) {
```
caarmen, line 424: Could we have some JsDoc? Even if it's not a public api, it could help the maintainers of this project to understand the purpose of this function, without reading all of the code. 🙏🏻

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+        const loader = type === 'front' ? this.frontLoader : this.backLoader;
+        const uploader = type === 'front' ? this.frontUploader : this.backUploader;
+        const fileInfo = type === 'front' ? this.frontFileInfo : this.backFileInfo;
+
+        try {
+            const isLargeFile = file && file.size > 10 * 1024 * 1024;
+            if (isLargeFile) {
+                loader.classList.remove('hidden');
+                uploader.classList.add('hidden');
+                fileInfo.classList.remove('hidden');
```
caarmen, line 438: I don't notice any visible difference with a small (3Mb file) and a large one (over 500Mb). I see the preview right away.

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+            let isImage = false;
+
+            if (file && file.type.startsWith('image/')) {
+                isImage = true;
+                mediaElement = new Image();
+                mediaElement.src = url;
+                await new Promise((resolve, reject) => {
+                    mediaElement.onload = resolve;
+                    mediaElement.onerror = reject;
+                });
```
caarmen, line 451: Could there ever be a race condition where the `onload` and `onerror` are set too late? If setting the `src` is processed so quickly, that by the time we do `mediaElement.onload` it's too late?

If so, perhaps the following could avoid it?
```javascript
                mediaElement = new Image();
                await new Promise((resolve, reject) => {
                    mediaElement.onload = resolve;
                    mediaElement.onerror = reject;
                    mediaElement.src = url;
                });

```

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+                }
+            }
+
+            if (type === 'front') {
+                this.frontMediaElement = mediaElement;
+            } else {
+                this.backMediaElement = mediaElement;
+            }
+
+            if (!isImage) {
```
caarmen, line 486: Can we have a comment here like `// Wait for video metadata (dimensions) to be available` ? (Or something similar).

Also is there not a constant for `1` for `readyState`? Or is widely known what `1` means, and I'm just not familiar with these apis? 😅

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+            const ctx = canvas.getContext('2d');
+
+            const drawFrame = () => {
+                if (sourceWidth && sourceHeight) {
+                    ctx.drawImage(mediaElement, 0, 0, sourceWidth, sourceHeight);
+                }
+            };
+
+            let audioTrack = null;
+            if (this.instance.options.microphone && !isImage) {
```
caarmen, line 530: Can we have a quick comment here explaining what this block is doing?

If I understand right, the scenario is:
* The user selected a video file for media injection
* The microphone is enabled

In this case we use the audio from the microphone, instead of from the video media file?

If I got this wrong, a comment is surely needed. 😅 

But I think, even if I got it right, a comment would be very helpful! 🙏🏻

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+                    }
+
+                    this.audioNodes = this.audioNodes || [];
+                    this.audioNodes.push({source, dest, type});
+                } catch (e) {
+                    log.error('[Camera] Web Audio API setup failed:', e);
+                }
+            }
+
+            let finalStream;
```
caarmen, line 558: A comment here would be welcome too, to explain what `finalStream` will be set set to.

I find it hard to follow with all the different streams (but once again, I'm not that familiar with this project).

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+
+            if (!isImage) {
+                try {
+                    await mediaElement.play();
+                } catch (e) {
+                    log.error('[Camera] Video play failed:', e);
+                }
+            }
+
+            if (isLargeFile && fileInfo.classList.contains('hidden')) {
```
caarmen: Not sure to understand this: we stop a file stream if it's a large file and the fileInfo isn't displayed? I don't understand.

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+                this.stopFileStream(type);
+                return;
+            }
+
+            const drawLoop = () => {
+                if (finalStream.active) {
+                    drawFrame();
+                    requestAnimationFrame(drawLoop);
+                }
+            };
```
caarmen: Is there any scenario where `finalStream.active` won't be set to `false`, but we'll want to stop the animation anyway?

If so, it seems we could do `this.animationFrameId = requestAnimationFrame(drawLoop);` and then, from the code which wants to interrupt this animation, call `cancelAnimationFrame(this.animationFrameId).`

caarmen: I saw earlier that we hardcoded `30` fps:

```javascript
const canvasStream = canvas.captureStream(30);
```

Is this loop being called at the same rate? If the loop is called more frequently than 30fps, wouldn't we be drawing too often?

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+        const loader = type === 'front' ? this.frontLoader : this.backLoader;
+        if (loader) {
+            loader.classList.add('hidden');
+        }
+
+        if (this.audioNodes) {
+            this.audioNodes = this.audioNodes.filter((node) => {
+                if (node.type === type) {
+                    try {
+                        node.source.disconnect();
```
caarmen, line 706: Should we be having side effects inside a `filter`? I think this isn't idiomatic for functional programming (methods like `map`, `filter`, etc typically should just return data, not have side effects).

Doing a filter first to get the relevant audioNodes, and then doing a for/forEach to `disconnect()` them would indeed end up processing the list twice.

But we're talking about just a few elements at most, right? Not sure this is a critical performance path. Clarity might be a better focus here in the clarity/perf tradeoff, no?

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+                return true;
+            });
+        }
+
+        const mediaElement = type === 'front' ? this.frontMediaElement : this.backMediaElement;
+        if (mediaElement) {
+            if (typeof mediaElement.pause === 'function') {
+                mediaElement.pause();
+            }
+            if (typeof mediaElement.load === 'function') {
```
caarmen, line 728: Why not:

```javascript
            mediaElement.src = '';
            if (typeof mediaElement.load === 'function') {
                mediaElement.load();
            }
```

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+                        log.warn('Error disconnecting audio node:', e);
+                    }
+                    return false;
+                }
+                return true;
+            });
+        }
+
+        const mediaElement = type === 'front' ? this.frontMediaElement : this.backMediaElement;
+        if (mediaElement) {
```
caarmen, line 722: Could we have a comment at the top of this block?

Basically, in functions which are kind of large (this one is around 70 lines), with several sections, it might help to browse the code if each section has a quick comment.

Another option would be to split the function into multiple smaller functions, each one doing only one very small thing. Then, comments might not be needed. Another tradeoff :)

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+                    t.stop();
+                });
+                this.backStream = null;
+            }
+            this.backVideo.classList.add('hidden');
+            this.backVideo.srcObject = null;
+            this.backFileInfo.classList.add('hidden');
+            this.backUploader.classList.remove('hidden');
+
+            this.instance.mediaManager.stopVideoStreaming('back');
```
caarmen: Quite a bit of copy/paste for the actions on the front/back streams. Not sure it would be simple to reduce this though. It might require more refactoring. Like extracting this into a function that takes a stream, video, fileinfo, and uploader arguments.

Then we could do something like:
```javascript
if (type == "front") {
    myFunctionWithAGoodName(
        this.frontStream,
        this.frontVideo,
        this.frontFileInfo,
        this.frontUploader,
        "front"
    };
} else {
    myFunctionWithAGoodName(
        this.backStream,
        this.backVideo,
        this.backFileInfo,
        this.backUploader,
        "back"
    };
}

myFunctionWithAGoodName(
    stream,
    video,
    fileInfo,
    uploader,
    someBetterName
){
    if (stream) {
        stream.getTracks().forEach((t) => {
            t.stop();
        }); 
        stream = null;
    }   
    video.classList.add('hidden');
    video.srcObject = null;
    fileInfo.classList.add('hidden');
    uploader.classList.remove('hidden');
    this.instance.mediaManager.stopVideoStreaming(someBetterName);
}
```

A way to avoid passing in so many arguments would be to define an object containing properties for all these objects.

src/plugins/Camera.js
```diff
@@ -60,4 +61,675 @@ export default class Camera extends OverlayPlugin {
...
+
+        if (deviceId === 'file') {
+            uploader.classList.remove('hidden');
+            await this.instance.mediaManager.stopVideoStreaming(type);
+
+            if (!this.frontStream && !this.backStream) {
+                this.toolbarBtn.setIndicator('');
+            }
+            return;
+        }
```
caarmen, line 813: Looks like the exact same logic for `deviceId === 'none'` and `deviceId === 'file'`. Did I miss something?

caarmen (commented):
I know it's in draft, but I had a couple of hours to look at this.

If any of my comments are irrelevant because you've been working on changes, just ignore them :) 

I basically just looked at two files today: 
* `GmDropdown.js`
* `Camera.js`

I'll take a look at other files when I get a chance.

src/plugins/Camera.js
caarmen: Nothing to see here! I had opened the html file directly, not from a web server.

src/plugins/Camera.js
caarmen: the first branch has placeholder, the second one has uploader

src/plugins/MediaManager.js
```diff
@@ -343,12 +392,25 @@ export default class MediaManager {
...
+                this.instance.peerConnection.removeTrack(sender);
+                if (type === 'front') {
+                    this.frontCameraSender = null;
+                } else {
+                    this.backCameraSender = null;
+                }
             }
-            // if the if the flag for bundled audio&video stream is set, we'll remove the audio track too
+            /*
+             * if the if the flag for bundled audio&video stream is set, we'll remove the audio track too
```
caarmen: if the if the flag => If the flag

tests/unit/camera.test.js
```diff
@@ -38,5 +49,68 @@ describe('Camera Plugin', () => {
             expect(tooltip.innerHTML).toEqual(expect.stringContaining('TEST CAMERA PLUGIN BUTTON TITLE'));
             document.querySelector('.gm-camera-button').parentElement.dispatchEvent(new Event('mouseleave'));
         });
+
+        test('widget is rendered with correct elements', () => {
+            // Check for dropdowns
+            expect(document.getElementsByClassName('gm-camera-dropdown-front')).toHaveLength(1);
+            /*
```
caarmen: Could you check this comment?

tests/unit/gappsInstall.test.js
```diff
@@ -40,7 +40,7 @@ describe('GAPPSInstall Plugin', () => {
             GAPPS_TITLE: 'TEST GAPPS TITLE',
             DRAG_DROP_TEXT: 'TEST DRAG DROP TEXT',
             BROWSE_BUTTON_TEXT: 'TEST BROWSE',
-            FILE_TYPE_NOT_APK: 'TEST FILE TYPE NOT APK',
+            FILE_TYPE_INVALID: 'TEST FILE TYPE NOT APK',
```
caarmen, line 43: Should the value be changed too?

package.json
```diff
@@ -8,7 +8,7 @@
     ],
     "types": "dist/index.d.ts",
     "engines": {
-        "node": ">=16"
+        "node": ">=20"
```
caarmen, line 11: This could be a breaking change for some applications using the player.

vite.config.js
jparez: No dependency to the hot reload, but browser grant access to camera only on https scheme

src/plugins/Camera.js
jparez: i had push a small commit with a small refacto of destroy flow.

Here is a summary of the changes:

- Lifecycle Management: DeviceRenderer.destroy() now dynamically iterates over all registered widgets and calls their destroy() method.
- AudioContext: The Camera plugin now implements a destroy() method that explicitly closes the AudioContext and removes event listeners.
- Store Cleanup: Added a destroy() method to the Store to centrally clear all subscriptions, preventing memory leaks from lingering listeners.
- General Cleanup: Also addressed minor leaks like unrevoked Blob URLs and dead code in the Factory.

So to answer the question: Yes, the AudioContext is now properly closed when the player is destroyed :)

src/plugins/Camera.js
jparez: see comment above

src/plugins/Camera.js
jparez: works on firefox, fixed on chrome

src/plugins/Camera.js
jparez: NotFoundError is standard, we will keep only this check this modern browser send this code

src/plugins/Camera.js
jparez: with my connection i can see the loader, i will add a screen on description section

src/plugins/Camera.js
jparez: await new Promise stop the code and wait for onload (or onerror) triggered, but even if there is no bug currently, your code is more readable, so let's do this.

src/plugins/Camera.js
jparez: no, we extracts the audio from the video file itself (if microphone is enable), Basically, we replace the physical microphone with the video's audio track. I will add a comment

src/plugins/Camera.js
jparez: It's to manage a race condition, if the file is large, the metadata can take a moment to be load 

```
await new Promise((resolve) => {
                    if (
                        mediaElement.readyState >= HTMLMediaElement.HAVE_METADATA &&
```
in this case if the user click on the cross to cancel the custom media stream (or change the source) we can have a issue when the stream is ready.

But check the DOM isn't robust, i will add an unique requestId and check when video is ready if the requestId is still revelant

tests/unit/gappsInstall.test.js
jparez: Since it's for test i18 it could, but in fact this is the right message

package.json
jparez: I had changed the logic to avoid force the engine for the users of the player and check node version for the developper

src/plugins/Camera.js
caarmen: This has been rewritten to use a requestVideoFrameCallback method when available, or falling back to a manual throttling.

In the case of manual throttling: we will write to and capture from the canvas at 30fps max (assuming requestAnimationFrame is at least 30fps).

In the case of requestVideoFrameCallback, if this supports, for example 60fps, then we will write to the canvas twice as much as we capture from it (we will miss some images). 

If the requestVideoFrameCallback supports only 15 fps, we will write to the canvas slower than we read from it (we will read the same image twice).

tests/unit/gappsInstall.test.js
caarmen: This is for a gapps test, so an apk message makes more sense

package.json
caarmen: I see it's mentioned in CONTRIBUTING.md, but it may be useful to note this either in the readme and/or the release notes.

vite.config.js
caarmen: Can we have a small comment explaining what we do here?

> // For development, instead of using the built dist, hot reload the files from src

src/plugins/Camera.js
caarmen: I tested on Chrome.

I'm able to select the same file multiple times: the video is replaced, but the previous audio continues to play at the same time as the new audio.

package.json
jparez: i revert this and assume player needs node 20 (release in 2018...). This will be a breaking change, indeed

src/plugins/Camera.js
caarmen: Tested the latest commit: now there's no audio preview for the video file, but that's fine! Probably better like this :)

I tested:
* Choose a video file.
* Choose the same video file multiple times.
* Open the camera app.
* Start to record a video.
* Stop the recording.
* Open the gallery app.
* Play the recorded video.
* => Only heard one audio track, looks good!

package.json
caarmen: Node 20 was released in 2023.

If we're able to support node 18 (released in 2022), it could offer a larger compatibility with other applications.

caarmen: Discussed with you offline: node 20 is required for vite.

caarmen: Ok the readme is up to date :)

PR #163, opened 2026-01-05
[PLAYER-92] refacto components

jparez:
## Description

Migrate legacy custom components to Web Components.

This PR refactors the legacy component pattern to native Web Components, improving maintainability and leveraging browser standards.

### New Web Components
- **GmTextInput**: Text input component with validation, min/max constraints, and error messaging

    - Replaces createTextInput factory
    - Supports regexFilter for input blocking and regexValidField for validation
    - Min/max validation with optional blocking via block-invalid attribute
    - Automatic error message generation from constraints

- **GmSlider**: Range slider component

    - Replaces createSlider factory
    - Dispatches gm-slider-input and gm-slider-change events

- **GmSwitch**: Toggle switch component

    - Replaces createSwitch factory
    - Dispatches gm-switch-change event

- **GmDropdown**: Select dropdown component

    - Replaces createDropDown factory
   - Supports dynamic option updates and keyboard navigation
 
- **GmChip**: Chip tag

### Updated Plugins

- **GPS**: Migrated to use <gm-text-input> for latitude, longitude, altitude, and speed inputs with min/max validation
- **Battery**: Migrated to use <gm-text-input> with blocking validation for charge level (0-100)
- **Network, IOThrottling, Identifiers, Phone, BasebandRIL**: Updated to use new web components

### Technical Details

- All components are registered as custom elements (e.g., customElements.define('gm-text-input', GmTextInput))
- Maintains backward compatibility with existing plugin APIs via legacy methods (getValue(), setValue(), etc.)
- Uses native HTML attributes and properties for configuration
- Custom events follow naming convention: gm-* (e.g., gm-input, gm-blur, gm-slider-change)

### Migration Benefits

- Standards-based: Uses native browser APIs instead of custom factories
- Encapsulation: Better component isolation and lifecycle management
- Declarative: Components can be used directly in HTML with attributes
- Maintainability: Clearer structure with class-based components
- Performance: Leverages native browser optimizations

### DIFF
--- a/tests/unit/fingerprint.test.js
+++ b/tests/unit/fingerprint.test.js
@@ -37,7 +37,7 @@ describe('FingerPrint Plugin', () => {
             const switchEl = document.querySelector('.autoValidationSwitch');
             expect(switchEl).toBeTruthy();
             expect(switchEl.checked).toBe(false);
-            switchEl.dispatchEvent(new CustomEvent('gm-switch-change', {detail: {checked: true}, bubbles: true}));
+            switchEl.click();
             expect(switchEl.checked).toBe(true);
 
             // body
@@ -79,9 +79,7 @@ describe('FingerPrint Plugin', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
             expect(sendEventSpy).toHaveBeenCalledTimes(0);
             const autoValidationSwitch = document.querySelector('.autoValidationSwitch');
-            autoValidationSwitch.dispatchEvent(
-                new CustomEvent('gm-switch-change', {detail: {checked: true}, bubbles: true}),
-            );
+            autoValidationSwitch.click();
 
             instance.emit('fingerprint', 'current_status scanning');
             instance.emit('fingerprint', 'scan start');
@@ -106,9 +104,7 @@ describe('FingerPrint Plugin', () => {
                         .querySelector('.gm-fingerprint-button')
                         .parentElement.classList.contains('gm-toolbar-dot-active'),
                 ).toBe(false);
-                autoValidation.dispatchEvent(
-                    new CustomEvent('gm-switch-change', {detail: {checked: true}, bubbles: true}),
-                );
+                autoValidation.click();
                 expect(
                     document
                         .querySelector('.gm-fingerprint-button')
--- a/tests/unit/iothrottling.test.js
+++ b/tests/unit/iothrottling.test.js
@@ -90,11 +90,9 @@ describe('IOThrottling Plugin', () => {
 
                 sendEventSpy.mockClear();
                 instance.outgoingMessages = [];
-                const dropDownProfile = diskio.profilesForDropdown.find((p) => p.value === profile.readByteRate);
-                diskio.dropdownProfile.value = dropDownProfile.value;
-                diskio.dropdownProfile.dispatchEvent(
-                    new CustomEvent('gm-dropdown-change', {detail: {value: dropDownProfile.value}, bubbles: true}),
-                );
+                const dropDownProfileIndex = diskio.profilesForDropdown
+                    .findIndex((p) => p.value === profile.readByteRate);
+                diskio.dropdownProfile.dropdownMenuDiv.children[dropDownProfileIndex].click();
                 diskio.widget.querySelector('.gm-btn').click();
                 expect(sendEventSpy).toHaveBeenCalledTimes(1);
                 expect(instance.outgoingMessages[0]).toEqual({
@@ -105,15 +103,12 @@ describe('IOThrottling Plugin', () => {
 
             sendEventSpy.mockClear();
             instance.outgoingMessages = [];
-            const dropDownProfile = diskio.profilesForDropdown.find((p) => p.value === 'Custom');
-            diskio.dropdownProfile.value = dropDownProfile.value;
-            diskio.dropdownProfile.dispatchEvent(
-                new CustomEvent('gm-dropdown-change', {detail: {value: dropDownProfile.value}, bubbles: true}),
-            );
-            diskio.readByteRate.value = 69;
-            diskio.readByteRate.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: 69}, bubbles: true}),
-            );
+            const dropDownProfileIndex = diskio.profilesForDropdown.findIndex((p) => p.value === 'Custom');
+            diskio.dropdownProfile.dropdownMenuDiv.children[dropDownProfileIndex].click();
+
+            const input = diskio.readByteRate.querySelector('input');
+            input.value = 69;
+            input.dispatchEvent(new Event('input', {bubbles: true}));
             diskio.widget.querySelector('.gm-btn').click();
             expect(sendEventSpy).toHaveBeenCalledTimes(1);
             expect(instance.outgoingMessages[0]).toEqual({
--- a/tests/unit/network.test.js
+++ b/tests/unit/network.test.js
@@ -131,20 +131,17 @@ describe('Network Plugin', () => {
             network = new Network(instance, {});
         });
 
-        test('wifi emit status event', () => {
+        test('wifi state', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
-            // Simulate gm-switch change to checked
-            network.wifiSwitch.dispatchEvent(
-                new CustomEvent('gm-switch-change', {detail: {checked: true}, bubbles: true}),
-            );
+            const wifiSwitch = network.wifiSwitch;
+            wifiSwitch.click();
             expect(sendEventSpy).toHaveBeenCalledWith({channel: 'settings', messages: ['enableif wifi']});
         });
 
-        test('mobile data emit status event', () => {
+        test('mobile data state', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
-            network.mobileDataSwitch.dispatchEvent(
-                new CustomEvent('gm-switch-change', {detail: {checked: true}, bubbles: true}),
-            );
+            const mobileDataSwitch = network.mobileDataSwitch;
+            mobileDataSwitch.click();
             expect(sendEventSpy).toHaveBeenCalledWith({channel: 'settings', messages: ['enableif mobile']});
             expect(sendEventSpy).toHaveBeenCalledWith({channel: 'network_profile', messages: ['notify phone']});
         });
@@ -152,27 +149,19 @@ describe('Network Plugin', () => {
         test('change network_profile emit event', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
             // change network profile
-            const dropDownSelectProfile = network.profilesForDropdownNetworkType.find((p) => p.value === 'lte');
-            network.dropdownNetworkType.value = dropDownSelectProfile.value;
-            network.dropdownNetworkType.dispatchEvent(
-                new CustomEvent('gm-dropdown-change', {detail: {value: dropDownSelectProfile.value}, bubbles: true}),
-            );
+            const dropDownSelectProfileIndex = network.profilesForDropdownNetworkType
+                .findIndex((p) => p.value === 'lte');
+            network.dropdownNetworkType.dropdownMenuDiv.children[dropDownSelectProfileIndex].click();
             expect(sendEventSpy).toHaveBeenCalledWith({
                 channel: 'network_profile',
                 messages: ['setprofile mobile lte'],
             });
 
             // change signal strength
-            const dropdownSignalStrengthProfile = network.profilesForDropdownSignalStrength.find(
+            const dropdownSignalStrengthProfileIndex = network.profilesForDropdownSignalStrength.findIndex(
                 (p) => p.value === 'great',
             );
-            network.selectMobileSignalStrength.value = dropdownSignalStrengthProfile.value;
-            network.selectMobileSignalStrength.dispatchEvent(
-                new CustomEvent('gm-dropdown-change', {
-                    detail: {value: dropdownSignalStrengthProfile.value},
-                    bubbles: true,
-                }),
-            );
+            network.selectMobileSignalStrength.dropdownMenuDiv.children[dropdownSignalStrengthProfileIndex].click();
             expect(sendEventSpy).toHaveBeenCalledWith({
                 channel: 'network_profile',
                 messages: ['setsignalstrength mobile great'],
--- a/tests/unit/phone.test.js
+++ b/tests/unit/phone.test.js
@@ -57,18 +57,16 @@ describe('Phone Plugin', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
 
             ['jean-michel', ''].forEach((invalidValue) => {
-                phone.phoneInput.value = invalidValue;
-                phone.phoneInput.dispatchEvent(
-                    new CustomEvent('gm-text-input-change', {detail: {value: invalidValue}, bubbles: true}),
-                );
+                const input = phone.phoneInput.querySelector('input');
+                input.value = invalidValue;
+                input.dispatchEvent(new Event('input', {bubbles: true}));
                 phone.phoneBtn.click();
                 expect(sendEventSpy).toHaveBeenCalledTimes(0);
             });
 
-            phone.phoneInput.value = '0123456789';
-            phone.phoneInput.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '0123456789'}, bubbles: true}),
-            );
+            const input = phone.phoneInput.querySelector('input');
+            input.value = '0123456789';
+            input.dispatchEvent(new Event('input', {bubbles: true}));
             phone.phoneBtn.click();
             expect(sendEventSpy).toHaveBeenCalledTimes(1);
 
@@ -77,10 +75,9 @@ describe('Phone Plugin', () => {
 
         test('sms send', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
-            phone.phoneInput.value = '0123456789';
-            phone.phoneInput.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '0123456789'}, bubbles: true}),
-            );
+            const input = phone.phoneInput.querySelector('input');
+            input.value = '0123456789';
+            input.dispatchEvent(new Event('input', {bubbles: true}));
 
             const event = new KeyboardEvent('keyup', {key: ''});
             phone.textInput.dispatchEvent(event);
--- a/tests/unit/baseband.test.js
+++ b/tests/unit/baseband.test.js
@@ -113,35 +113,18 @@ describe('BasebandRIL Plugin', () => {
             instance.emit('baseband', 'network operator 123456');
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
 
-            baseband.networkOperatorMMC.value = '123456';
-            baseband.networkOperatorMMC.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '123456'}, bubbles: true}),
-            );
+            const setInput = (component, value) => {
+                const input = component.querySelector('input');
+                input.value = value;
+                input.dispatchEvent(new Event('input', {bubbles: true}));
+            };
 
-            baseband.networkOperatorName.value = 'value';
-            baseband.networkOperatorName.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: 'value'}, bubbles: true}),
-            );
-
-            baseband.simOperatorMMC.value = '123456';
-            baseband.simOperatorMMC.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '123456'}, bubbles: true}),
-            );
-
-            baseband.simOperatorName.value = 'value';
-            baseband.simOperatorName.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: 'value'}, bubbles: true}),
-            );
-
-            baseband.simMSIN.value = '012345678';
-            baseband.simMSIN.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '012345678'}, bubbles: true}),
-            );
-
-            baseband.simOperatorPhoneNumber.value = '0011223344';
-            baseband.simOperatorPhoneNumber.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '0011223344'}, bubbles: true}),
-            );
+            setInput(baseband.networkOperatorMMC, '123456');
+            setInput(baseband.networkOperatorName, 'value');
+            setInput(baseband.simOperatorMMC, '123456');
+            setInput(baseband.simOperatorName, 'value');
+            setInput(baseband.simMSIN, '012345678');
+            setInput(baseband.simOperatorPhoneNumber, '0011223344');
             baseband.submitBtn.click();
 
             expect(sendEventSpy).toHaveBeenCalledTimes(1);
@@ -157,9 +140,9 @@ describe('BasebandRIL Plugin', () => {
                 ],
             });
 
-            baseband.simOperatorName.value = 'value';
-            baseband.simMSIN.value = '012345678';
-            baseband.simOperatorPhoneNumber.value = '0011223344';
+            setInput(baseband.simOperatorName, 'value');
+            setInput(baseband.simMSIN, '012345678');
+            setInput(baseband.simOperatorPhoneNumber, '0011223344');
             baseband.submitBtn.click();
             expect(sendEventSpy).toHaveBeenCalledTimes(2);
             expect(instance.outgoingMessages[1]).toEqual({
--- a/tests/unit/gps.test.js
+++ b/tests/unit/gps.test.js
@@ -69,10 +69,10 @@ describe('GPS Plugin', () => {
                 test(`button disabled when ${field} is invalid`, () => {
                     const fieldInput = gps.inputComponents[field];
                     const submitButton = document.querySelector('.gm-gps-update');
-                    fieldInput.value = value;
-                    fieldInput.dispatchEvent(
-                        new CustomEvent('gm-text-input-change', {detail: {value: value}, bubbles: true}),
-                    );
+
+                    const input = fieldInput.querySelector('input');
+                    input.value = value;
+                    input.dispatchEvent(new Event('input', {bubbles: true}));
                     if (field !== 'accuracy' && field !== 'bearing') {
                         expect(
                             fieldInput.querySelector('.text-input-message').classList.contains('hidden'),
@@ -109,63 +109,13 @@ describe('GPS Plugin', () => {
             sendEventSpy.mockRestore();
         });
 
-        /*
-         *test('invalid input value', () => {
-         *    gps.inputComponents.altitude.value = 'jean-michel';
-         *    gps.inputComponents.altitude.dispatchEvent(new CustomEvent('gm-text-input-change', { detail: { value: 'jean-michel' }, bubbles: true }));
-         *
-         *    gps.inputComponents.latitude.value = 'jean-michel';
-         *    gps.inputComponents.latitude.dispatchEvent(new CustomEvent('gm-text-input-change', { detail: { value: 'jean-michel' }, bubbles: true }));
-         *
-         *    gps.inputComponents.longitude.value = 'jean-michel';
-         *    gps.inputComponents.longitude.dispatchEvent(new CustomEvent('gm-text-input-change', { detail: { value: 'jean-michel' }, bubbles: true }));
-         *
-         *    gps.inputComponents.accuracy.value = 'jean-michel';
-         *    gps.inputComponents.accuracy.dispatchEvent(new CustomEvent('gm-text-input-change', { detail: { value: 'jean-michel' }, bubbles: true }));
-         *
-         *    gps.inputComponents.bearing.value = 'jean-michel';
-         *    gps.inputComponents.bearing.dispatchEvent(new CustomEvent('gm-text-input-change', { detail: { value: 'jean-michel' }, bubbles: true }));
-         *
-         *    gps.inputComponents.speed.value = 'jean-michel';
-         *    gps.inputComponents.speed.dispatchEvent(new CustomEvent('gm-text-input-change', { detail: { value: 'jean-michel' }, bubbles: true }));
-         *    expect(document.querySelector('.gm-gps-update').disabled).toBeTruthy();
-         *
-         *    document.querySelector('.gm-gps-update').click();
-         *
-         *    expect(sendEventSpy).toHaveBeenCalledTimes(0);
-         *});
-         */
-
         test('min value', () => {
-            gps.inputComponents.altitude.value = '-10000';
-            gps.inputComponents.altitude.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '-10000'}, bubbles: true}),
-            );
-
-            gps.inputComponents.latitude.value = '-90';
-            gps.inputComponents.latitude.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '-90'}, bubbles: true}),
-            );
-
-            gps.inputComponents.longitude.value = '-180';
-            gps.inputComponents.longitude.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '-180'}, bubbles: true}),
-            );
-
-            gps.inputComponents.accuracy.value = '0';
-            gps.inputComponents.accuracy.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '0'}, bubbles: true}),
-            );
-
-            gps.inputComponents.bearing.value = '0';
-            gps.inputComponents.bearing.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '0'}, bubbles: true}),
-            );
-
-            gps.inputComponents.speed.value = '0';
-            gps.inputComponents.speed.dispatchEvent(
-                new CustomEvent('gm-text-input-change', {detail: {value: '0'}, bubbles: true}),
-            );
+            ['altitude', 'latitude', 'longitude', 'accuracy', 'bearing', 'speed'].forEach((field, index) => {
+                const vals = ['-10000', '-90', '-180', '0', '0', '0'];
+                const input = gps.inputComponents[field].querySelector('input');
+                input.value = vals[index];
+                input.dispatchEvent(new Event('input', {bubbles: true}));
+            });
             document.querySelector('.gm-gps-update').click();
 
             expect(sendEventSpy).toHaveBeenCalledTimes(1);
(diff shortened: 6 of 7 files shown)


src/plugins/GAPPSInstall.js
```diff
@@ -395,6 +392,8 @@ class InitialView {
                 log.error(error);
                 this.plugin.instance.store.dispatch({type: 'DRAG_AND_DROP_UPLOAD_FILE_ENABLED', payload: false});
             }
+        } else {
+            this.plugin.instance.store.dispatch({type: 'DRAG_AND_DROP_UPLOAD_FILE_ENABLED', payload: false});
```
caarmen, line 396: How does this additional line relate to the migration of the component?

src/components/GmSwitch.js
```diff
@@ -0,0 +1,106 @@
...
+        this.appendChild(input);
+        this.appendChild(slider);
+    }
+
+    /**
+     * Handles click events to toggle state.
+     * @param {Event} e - Click event.
+     */
+    #handleClick(e) {
+        e.preventDefault(); // Prevent double toggle if clicking label/input
```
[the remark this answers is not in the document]
jparez: The decision to keep `e.preventDefault()` is deliberate to ensuring robustness.

A solution will be to manage this by css rules, but since the component runs in Light DOM, relying solely on CSS (`pointer-events: none` on the input) to prevent double-toggling is fragile, as it could be easily overridden by the host application's styles.
Using `e.stopPropagation()` would be an anti-pattern here, breaking event bubbling expectations for parent containers. 
Therefore, `e.preventDefault()` remains the safest way to guarantee a single source of truth for the toggle state, regardless of the integration context.

src/plugins/Battery.js
```diff
@@ -151,55 +158,71 @@ export default class Battery extends OverlayPlugin {
...
+                this.updateUIBatteryChargingPercent(this.chargeInput.value);
+                this.sendDataToInstance();
             } else if (e.key === 'ArrowDown') {
                 e.preventDefault();
-                this.chargeInput.setValue(Math.max(0, Number(this.chargeInput.getValue()) - 1), true);
+                this.chargeInput.value = Math.max(0, Number(this.chargeInput.value) - 1);
+                // Manually trigger updates
+                this.chargeSlider.value = parseFloat(this.chargeInput.value);
+                this.updateUIBatteryChargingPercent(this.chargeInput.value);
+                this.sendDataToInstance();
```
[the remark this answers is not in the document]
jparez: While silence on programmatic changes is standard native behavior, as a 'smart' component, we introduced 
setValue(val, {emit: true}). This allows forcing event emission when needed (like programmatic arrow-key navigation), simplifying the synchronization logic in consumers like Battery.js.
This pattern is commonly found in modern frameworks.

src/components/GmTextInput.js
```diff
@@ -0,0 +1,410 @@
...
+         * Avoid duplicate events if possible, but duplicate change is usually harmless for validation.
+         */
+        e.stopPropagation(); // Stop native change bubbling
+        this.dispatchEvent(
+            new CustomEvent('gm-text-input-change', {
+                detail: {value: this.#value},
+                bubbles: true,
+            }),
+        );
+    }
```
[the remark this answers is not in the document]
jparez: The duplicate event dispatch is intentional by design to support real-time validation and UI feedback (e.g., updating map markers or checking regex validity while typing) using a single event listener in consumer plugins. Refactoring consumers to listen to separate input and change events would be a **significant breaking change** across the entire plugin ecosystem. The side effects in listeners are idempotent (UI updates/validations), making the redundancy harmless in this specific context

src/plugins/Battery.js
```diff
@@ -151,55 +158,71 @@ export default class Battery extends OverlayPlugin {
...
+            this.sendDataToInstance();
+        });
+
+        // Handle blur to reset empty value to 0
+        this.chargeInput.inputElement?.addEventListener('blur', (e) => {
+            if (e.target.value === '') {
+                this.chargeInput.value = '0';
+                // Trigger change manually if needed, or rely on bind
+            }
         });
```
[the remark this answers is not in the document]
jparez: useless code since the listener `this.chargeInput.addEventListener('gm-text-input-change'` check and set to 0 if parseFloat fail `parseFloat(value) || 0;`
So this code can be removed

src/components/GmTextInput.js
```diff
@@ -0,0 +1,410 @@
...
+            if (!isNaN(min) && min >= 0 && val < 0) {
+                e.target.value = this.#value;
+                return;
+            }
+            // Block values greater than max
+            if (!isNaN(max) && val > max) {
+                e.target.value = this.#value;
+                return;
+            }
+        }
```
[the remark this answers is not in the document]
jparez: This two validation mechaims does'nt have the same behavior : 

The regex block invalid character while strict-range check the semantic logic (if max number is 10 and user have type '1' then refuse > 0 value)

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+        if (this.selectedValueDiv) {
+            let display = newValue;
+            if (item) {
+                if (typeof item === 'string') {
+                    display = item;
+                } else {
+                    display = item.valueToDisplay ?? item.element?.innerHTML ?? item.value;
+                }
+            }
+            this.selectedValueDiv.innerHTML = display || 'Select...';
```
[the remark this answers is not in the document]
jparez: The input data comes from trusted static configurations, not arbitrary user input, making the XSS risk negligible in this context.. Since we want to support rich content rendering (like icons) in dropdown options we will keep innerHTML

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+            return;
+        }
+        this.dropdownMenuDiv.innerHTML = '';
+
+        this.#items.forEach((item) => {
+            const optionDiv = document.createElement('div');
+            optionDiv.className = 'dropdown-item';
+
+            if (typeof item === 'string') {
+                optionDiv.innerHTML = item;
```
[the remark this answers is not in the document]
jparez: Same that the comment above

src/components/GmSwitch.js
```diff
@@ -0,0 +1,106 @@
...
+            this.#render();
+        }
+        this.addEventListener('click', this.#handleClick.bind(this));
+    }
+
+    /**
+     * Called when custom element is removed from the DOM.
+     */
+    disconnectedCallback() {
+        this.removeEventListener('click', this.#handleClick.bind(this));
```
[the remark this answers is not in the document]
jparez: Oups, this is memory leak. We need to keep the ref to the handleClick fn, since bond create a new ref in memory

src/components/GmProgressBar.js
```diff
@@ -0,0 +1,117 @@
...
+     * Set current value.
+     * Clamps value between 0 and max.
+     * @param {number} newValue - The new value.
+     */
+    set value(newValue) {
+        const val = Math.max(0, Math.min(newValue, this.#max));
+        this.setAttribute('value', val);
+    }
+
+    /**
```
caarmen, line 67: It seems that the max property doesn't need to be exposed at this point.Maybe in the future it might be useful, but currently it's not used by other code in the project.

src/components/GmSlider.js
```diff
@@ -0,0 +1,221 @@
...
+            if (this.sliderCursor) {
+                this.sliderCursor.style.left = adjustedPercentage;
+                const calc = ((percentage / 100) * cursorWidth - cursorWidth / 2) * -1;
+                this.sliderCursor.style.transform = `translate(${calc}px, -50%)`;
+            }
+        });
+    }
+
+    // Expose min/max as properties to support direct assignments (e.g., el.min = 0)
+    get min() {
```
caarmen, line 187: Seems like min/max properties aren't currently used by users of the component.

src/components/GmSlider.js
```diff
@@ -0,0 +1,221 @@
+/**
+ * Custom Slider Component
+ * Usage: <gm-slider min="0" max="100" value="50"></gm-slider>
```
caarmen, line 3: Add onCursorMove?

src/components/GmSlider.js
```diff
@@ -0,0 +1,221 @@
...
+    static get observedAttributes() {
+        return ['min', 'max', 'value', 'disabled'];
+    }
+
+    constructor() {
+        super();
+    }
+
+    connectedCallback() {
+        if (!this.querySelector('.slider-input')) {
```
caarmen, line 20: In which case would we not have the `.slider-input`?

If this could happen, maybe the component wouldn't work at all?

caarmen: You just explained to me: if we find the slider-input element, this means we already did the render, and don't have to do it again.

Maybe add comment or more explicit code (a field to note if we already did the render?)

src/components/GmSlider.js
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        const val = parseFloat(newValue);
+        // For 'disabled', newValue can be null or an empty string, which parseFloat would make NaN.
+        if (isNaN(val) && name !== 'disabled') {
+            return;
+        }
+
+        switch (name) {
+            case 'min':
```
caarmen, line 38: The logic here for min/max is a new feature currently not used anywhere.


UPDATE: in fact, it IS used: we construct the element with no attributes, and then we set all the attributes separately (min, max, value).

caarmen: Maybe need a check that the new min isn't greater than the current max

src/components/GmSlider.js
```diff
@@ -0,0 +1,221 @@
...
+                this.#min = val;
+                if (this.input) {
```
caarmen, line 48: Check if it's really possible for `this.input` to not have been created already.

src/components/GmSlider.js
```diff
@@ -0,0 +1,221 @@
...
+                // Validation for max: ensure it's not less than or equal to min
+                if (val <= this.#min) {
+                    // If invalid, revert the attribute to the old value or ignore.
+                    return;
+                }
+                this.#max = val;
+                if (this.input) {
+                    this.input.max = this.#max;
+                    // Validate value against new max
+                    if (this.#value > this.#max) {
```
caarmen: We adjust the value to ensure it's within the bounds, but only if `this.input` exists. Perhaps this adjustment to value should be done regardless of whether `this.input` exists.

Same for min.

src/plugins/Battery.js
```diff
@@ -38,12 +40,16 @@ export default class Battery extends OverlayPlugin {
...
-            this.chargingInput.setState(values[0] !== 'discharging');
+            this.chargingInput.checked = values[0] !== 'discharging';
 
             this.updateUIBatteryChargingState();
 
-            this.chargeSlider.setValue(values[1], false);
-            this.chargeInput.setValue(values[1]);
+            const numValue = parseFloat(values[1]);
+            // Only update slider if value is valid number, otherwise keep current value
+            if (!Number.isNaN(numValue)) {
```
caarmen: Check if this isNaN is necessary. If so, should it be in the component? (Since we have other checks in the component, like assuring that the min < max)

src/components/GmSlider.js
caarmen: Seen with you: in fact, the onCursorMove could be removed, because we now use events instead of callbacks.

src/components/GmSwitch.js
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Called when an observed attribute changes.
+     * @param {string} name - Attribute name.
+     * @param {string} oldValue - Old attribute value.
+     * @param {string} newValue - New attribute value.
+     */
+    attributeChangedCallback(name, oldValue, newValue) {
+        if (name === 'checked') {
+            this.checked = newValue !== null;
```
caarmen: this.#checked

src/components/GmSwitch.js
```diff
@@ -0,0 +1,114 @@
...
+        this.appendChild(slider);
+    }
+
+    /**
+     * Handles click events to toggle state.
+     * @param {Event} e - Click event.
+     */
+    #handleClick(e) {
+        e.preventDefault(); // Prevent double toggle if clicking label/input
+        this.checked = !this.checked;
```
caarmen, line 123: this.#checked = !this.#checked

tests/unit/fingerprint.test.js
```diff
@@ -34,13 +34,11 @@ describe('FingerPrint Plugin', () => {
...
-            const firstSpan = document.querySelector('.autoValidationSwitch span');
-
-            expect(checkbox).toBeTruthy();
-            expect(checkbox.checked).toBe(false);
-            firstSpan.click();
-            expect(checkbox.checked).toBe(true);
+            const switchEl = document.querySelector('.autoValidationSwitch');
+            expect(switchEl).toBeTruthy();
+            expect(switchEl.checked).toBe(false);
+            switchEl.dispatchEvent(new CustomEvent('gm-switch-change', {detail: {checked: true}, bubbles: true}));
```
caarmen: why dispatch the event instead of click?

Should click and verify that we get the expected event.

tests/unit/fingerprint.test.js
```diff
@@ -80,8 +78,10 @@ describe('FingerPrint Plugin', () => {
         test("Auto validate fingerprint's request when auto validation is enabled", () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
             expect(sendEventSpy).toHaveBeenCalledTimes(0);
-            const autoValidation = document.querySelector('.autoValidationSwitch span');
-            autoValidation.click();
+            const autoValidationSwitch = document.querySelector('.autoValidationSwitch');
```
caarmen, line 81: Same comment as above (multiple places)

tests/unit/network.test.js
```diff
@@ -133,15 +133,17 @@ describe('Network Plugin', () => {
 
         test('wifi emit status event', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
-            network.wifiSwitch.setState(true, true);
-            expect(sendEventSpy).toHaveBeenCalledWith({channel: 'settings', messages: ['enableif wifi']});
+            network.wifiSwitch.checked = true;
+            network.wifiSwitch.dispatchEvent(new CustomEvent('gm-switch-change', { detail: { checked: true } }));
```
caarmen: Why do we need to explicitly dispatch the event here?

Same comment below.

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        // Selected Value Display
+        this.selectedValueDiv = document.createElement('div');
+        this.selectedValueDiv.className = 'dropdown-selected';
+        this.selectedValueDiv.textContent = this.#getDisplayValue(this.#value) || 'Select...';
+        this.appendChild(this.selectedValueDiv);
+
+        // Menu container
+        this.dropdownMenuDiv = document.createElement('div');
```
caarmen, line 41: Could it be possible to use a `select` element instead of `div`?

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+        // Event Listeners
+        this.selectedValueDiv.addEventListener('click', (e) => {
+            if (this.#disabled) {
+                return;
+            }
+            e.stopPropagation();
+            this.#toggleMenu();
+        });
+
+        document.addEventListener('click', (e) => {
```
caarmen, line 58: // Close the dropdown if the user clicks outside of it

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+     * @returns {Array} List of items.
+     */
+    get items() {
+        return this.#items;
+    }
+
+    /**
+     * Set the list of items.
+     * Re-renders the options and clears the value if it's no longer present.
+     * @param {Array} newItems - List of items.
```
caarmen: Can you document what are supported item types?

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+            return;
+        }
+
+        this.classList.add('dropdown');
+        if (this.#disabled) {
+            this.classList.add('disabled');
+        }
+
+        // Selected Value Display
+        this.selectedValueDiv = document.createElement('div');
```
caarmen, line 35: Maybe put this part in a `#render()` method?

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+     * Set the list of items.
+     * Re-renders the options and clears the value if it's no longer present.
+     * @param {Array} newItems - List of items.
+     */
+    set items(newItems) {
+        this.#items = newItems;
+        if (this.dropdownMenuDiv) {
+            this.#renderOptions();
+            const found = this.#items.some((i) => (i.value ?? i) === this.#value);
+            if (!found && this.#items.length > 0) {
```
caarmen: Could simplify this and remove `this.#items.length > 0` (it's not tested, and shouldn't happen, and if it DID happen, we don't know if the component may be broken somewhere else).

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+                this.#closeMenu();
+            } else {
+                this.classList.remove('disabled');
+            }
+        }
+    }
+
+    /**
+     * Helper to retrieve the display label for a given value.
+     * @param {string|number} val - The value to look for.
```
caarmen: I think only string is supported.

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+     * @returns {string} The display text.
+     */
+    #getDisplayValue(val) {
+        if (!this.#items) {
+            return val;
+        }
+        const item = this.#items.find((i) => (i.value ?? i) === val);
+        if (!item) {
+            return val;
+        }
```
caarmen, line 166: Is this correct? If we don't find the item, we return the `val` passed in argument?

src/components/GmDropdown.js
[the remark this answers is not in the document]
caarmen: discussed in pair, maybe use textContent instead.

src/components/GmDropdown.js
```diff
@@ -0,0 +1,311 @@
...
+                /*
+                 * If item is string, we don't have reference to DOM element easily unless we stored it.
+                 * Fallback: childNodes array.
+                 */
+                parent = this.dropdownMenuDiv.children[index];
+            }
+
+            if (!parent) {
+                return;
+            }
```
caarmen: In which case could we not have a parent?

src/plugins/Network.js
```diff
@@ -114,12 +114,12 @@ export default class Network extends OverlayPlugin {
...
         // Update the dropdowns network type and signal strength
-        this.dropdownNetworkType.setValue(
-            this.profilesForDropdownNetworkType.find((mp) => mp.value === mobileProfile[1]),
-        );
-        this.selectMobileSignalStrength.setValue(
-            this.profilesForDropdownSignalStrength.find((ms) => ms.value === signalStrength[1]),
-        );
+        this.dropdownNetworkType.value = this.profilesForDropdownNetworkType.find(
+            (mp) => mp.value === mobileProfile[1],
+        )?.value;
```
caarmen, line 119: Do we need `?.value` ?

tests/unit/iothrottling.test.js
```diff
@@ -91,7 +91,10 @@ describe('IOThrottling Plugin', () => {
                 sendEventSpy.mockClear();
                 instance.outgoingMessages = [];
                 const dropDownProfile = diskio.profilesForDropdown.find((p) => p.value === profile.readByteRate);
-                diskio.dropdownProfile.setValue(dropDownProfile, true);
+                diskio.dropdownProfile.value = dropDownProfile.value;
+                diskio.dropdownProfile.dispatchEvent(
```
caarmen: Should we do a click instead of dispatching the event from the test?

(Other places in this file too).

tests/unit/network.test.js
```diff
@@ -133,24 +133,30 @@ describe('Network Plugin', () => {
 
         test('wifi emit status event', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
-            network.wifiSwitch.checked = true;
-            network.wifiSwitch.dispatchEvent(new CustomEvent('gm-switch-change', { detail: { checked: true } }));
-            expect(sendEventSpy).toHaveBeenCalledWith({ channel: 'settings', messages: ['enableif wifi'] });
+            // Simulate gm-switch change to checked
```
caarmen: Should we be clicking to toggle the switch? (Maybe other locations in this file too)

src/components/GmChip.js
```diff
@@ -0,0 +1,145 @@
...
+     * @param {string} oldValue - Old attribute value.
+     * @param {string} newValue - New attribute value.
+     */
+    attributeChangedCallback(name, oldValue, newValue) {
+        if (oldValue === newValue) {
+            return;
+        }
+
+        if (name === 'type') {
+            this.type = newValue;
```
caarmen, line 45: this.#type

src/components/GmChip.js
```diff
@@ -0,0 +1,145 @@
...
+        } else if (name === 'value') {
+            this.value = newValue;
```
caarmen, line 47: this.#value

src/components/GmChip.js
```diff
@@ -0,0 +1,145 @@
...
+        if (this.#type === newType) {
+            return;
+        }
+
+        // Remove old class
+        if (this.#type) {
+            this.classList.remove('gm-tag-' + this.#type);
+        }
+
+        this.#type = newType || 'success';
```
caarmen: falling back to `success` might hide a bug where the caller calls `component.type=undefined`

src/components/GmChip.js
```diff
@@ -0,0 +1,145 @@
...
+        this.classList.add('gm-tag-' + this.#type);
+
+        // Reflect to attribute
+        if (this.getAttribute('type') !== this.#type) {
+            this.setAttribute('type', this.#type);
+        }
```
caarmen, line 74: Is this `setAttribute()` necessary?

caarmen: Same for setValue

src/components/GmChip.js
```diff
@@ -0,0 +1,145 @@
...
+        const container = document.createElement('div');
+        container.className = 'gm-tag-container';
+        container.textContent = this.#value;
+        this.appendChild(container);
+    }
+
+    /**
+     * Updates the CSS class based on type.
+     */
+    #updateTypeClass() {
```
caarmen: Maybe put this directly in `#render`

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+    attributeChangedCallback(name, oldValue, newValue) {
+        if (oldValue === newValue) {
+            return;
+        }
+
+        switch (name) {
+            case 'value':
+                this.value = newValue; // Goes through setter
+                break;
+            case 'regex-filter':
```
caarmen, line 59: Seems we can remove regex-filter, because the number ranges fill this need.

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+            const val = parseFloat(strValue);
+
+            // Block invalid number (NaN)
+            if (isNaN(val)) {
+                log.warn('GmTextInput: Ignored NaN value set programmatically (strict-range):', strValue);
+                return;
+            }
+
+            // Block negative input if min is non-negative
+            if (!isNaN(min) && min >= 0 && val < 0) {
```
caarmen: Maybe we can replace `min >= 0 && val < 0` by `val < min`

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+                log.warn('GmTextInput: Ignored negative value set programmatically (strict-range):', strValue);
```
caarmen: The error message will probably be adjusted too.

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+     * @param {boolean} [options.emit] - Whether to dispatch 'change' and 'input' events. Default false.
+     */
+    setValue(newValue, {emit = false} = {}) {
+        const strValue = String(newValue ?? '');
+        if (this.#value === strValue) {
+            return;
+        }
+
+        // Respect regex-filter on programmatic set (restore legacy behavior)
+        if (this.#regexFilter && !this.#regexFilter.test(strValue) && strValue !== '') {
```
caarmen, line 119: The regex filter isn't needed because we have min/max range

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+            }
+        }
+
+        this.#value = strValue;
+        if (this.input) {
+            this.input.value = this.#value;
+        }
+
+        if (emit) {
+            this.dispatchEvent(
```
caarmen, line 154: Seems like this first event can be removed?

Not sure what the difference is between the two events.

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+            this.input.readOnly = value !== null;
+        }
+    }
+
+    /**
+     * Updates specific UI elements based on attributes.
+     * @param {string} name - Attribute name.
+     * @param {string} value - Attribute value.
+     */
+    #updateUI(name, value) {
```
caarmen: Could we perhaps have two methods like `#updateAppendText()` and `#updateUnitText()`, instead of a generic `updateUi()`

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+                this.unitTextEl.classList.remove('hidden');
+            }
+        }
+        this.#updateBottomVisibility();
+    }
+
+    /**
+     * Updates the visibility of the bottom container.
+     */
+    #updateBottomVisibility() {
```
caarmen, line 346: Maybe this could go inside updateErrorMessage()? It's only used there.

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+    /**
+     * Handles input events from the internal input element.
+     * @param {Event} e - Input event.
+     */
+    #handleInput(e) {
+        const newValue = e.target.value;
+        const selectionStart = e.target.selectionStart;
+
+        // Regex Filter logic: revert if not matching
+        if (this.#regexFilter && !this.#regexFilter.test(newValue) && newValue !== '') {
```
caarmen, line 363: This will probably be removed since we'll remove the regexFilter

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+         * Strict Range Enforcement (prevent typing beyond limits)
+         * Only apply if 'strict-range' attribute is present
+         */
+        if (this.hasAttribute('strict-range') && this.getAttribute('type') === 'number' && newValue !== '') {
+            const min = parseFloat(this.getAttribute('min'));
+            const max = parseFloat(this.getAttribute('max'));
+            const val = parseFloat(newValue);
+
+            // Block negative input if min is non-negative
+            if (!isNaN(min) && min >= 0 && val < 0) {
```
caarmen: Seems like it may be duplicate of the logic in setValue.

Also: should check val < min

src/components/GmTextInput.js
```diff
@@ -0,0 +1,430 @@
...
+            // Block values greater than max
+            if (!isNaN(max) && val > max) {
+                e.target.value = this.#value;
+                return;
+            }
+        }
+
+        this.#value = newValue;
+
+        this.dispatchEvent(
```
caarmen, line 402: Probably will remove this event (as noted in setValue())

tests/unit/phone.test.js
```diff
@@ -57,12 +57,18 @@ describe('Phone Plugin', () => {
             const sendEventSpy = vi.spyOn(instance, 'sendEvent');
 
             ['jean-michel', ''].forEach((invalidValue) => {
-                phone.phoneInput.setValue(invalidValue, true);
+                phone.phoneInput.value = invalidValue;
```
caarmen: Maybe the test should be typing text, instead of programmatically setting the text on the component + dispatching the event? (Similar to what we saw in other tests).

src/plugins/Battery.js
```diff
@@ -195,30 +195,16 @@ export default class Battery extends OverlayPlugin {
...
         // bind arrow keys to input, to increase/decrease value with arrow up/down
         this.instance.addListener(this.chargeInput, 'keydown', (e) => {
             if (e.key === 'ArrowUp') {
                 e.preventDefault();
-                this.chargeInput.value = Math.min(100, Number(this.chargeInput.value) + 1);
-                // Manually trigger updates as setter doesn't emit events
-                this.chargeSlider.value = parseFloat(this.chargeInput.value);
-                this.updateUIBatteryChargingPercent(this.chargeInput.value);
-                this.sendDataToInstance();
+                const newValue = Math.min(100, Number(this.chargeInput.value) + 1);
```
caarmen: Can we just do `const newValue = Number(this.chargeInput.value) + 1` (without the `Math.min`)? With the min/max bounds now handled by the component, perhaps we don't need to manage the bounds here too?

src/plugins/Battery.js
```diff
@@ -195,30 +195,16 @@ export default class Battery extends OverlayPlugin {
             this.sendDataToInstance();
         });
 
-        // Handle blur to reset empty value to 0
-        this.chargeInput.inputElement?.addEventListener('blur', (e) => {
```
caarmen: Looks like we lost the blur handling here, but added it back in #159 ?

src/plugins/GAPPSInstall.js
jparez: If this.plugin.instance.options.fileUploadUrl is not provided but this.plugin.instance.options.fileUpload is set to true, the code will crash. Keep the code as is to safely handle the missing URL case without crashes or error logs.

src/components/GmProgressBar.js
jparez: Keeping the max property is the correct approach for a reusable web component for several reasons:

- Flexibility: It allows the progress bar to be used for values other than percentages or for use an other value than 100 for ma.
- Standard Compliance: It mimics the standard <progress> HTML element API, making it intuitive for other developers to use.

src/components/GmSlider.js
jparez: same comments than above

src/components/GmSlider.js
jparez: Yes, it is possible for this.input to be undefined at that point!

This happens due to the Web Component lifecycle:

connectedCallback (and so render()) is call only when the element is added to the DOM, not when the element is create, so in this case :

```
const slider = document.createElement('gm-slider');
slider.min = 0;
document.body.appendChild(slider); 
```

attributeChangedCallback (slider.min = 0) will be call before the render (called for document.body.appendChild(slider))

src/plugins/Battery.js
jparez: is not, the component already handle this case

src/components/GmSwitch.js
jparez: this.checked is syntaxtic sugar for this.checked(true). We call the setter fn 'checked' not the private member #checked.

BTW there was an error with the disable member in attributeChangedCallback, the var was not handle !

src/components/GmSwitch.js
jparez: nop, this is like `this.checked(!this.checked())` where `this.checked =` is  `set checked()` and `!checked()` is  `get checked()` 

Same that the comment above

src/components/GmDropdown.js
jparez: Since we need rich text inner the select options we cannot use a <select>. Sync a native <select> with our custom select would be overly complex for a limited benefit

src/components/GmDropdown.js
jparez: it's a fallback, in worst case we display the value. It's a safeguard to prevent an empty UI in case of temporary desynchronization or data errors.

src/components/GmDropdown.js
jparez: In case of developpment error, but this could mask a bug, i will remove this

src/plugins/Network.js
jparez: yes 

```
this.profilesForDropdownNetworkType = arr
            .sort((profA, profB) => profA.id - profB.id)
            .map((profile) => {
                const divContainer = document.createElement('div');
                const label = document.createElement('div');
                const icon = document.createElement('div');
                icon.className = 'gm-network-profile-icon ' + profile.icon;
                divContainer.appendChild(icon);
                divContainer.appendChild(label);
                divContainer.style.display = 'flex';
                divContainer.style.alignItems = 'center';
                label.innerHTML = profile.label || '';
                return {
                    element: divContainer,
                    value: profile.name,
                    valueToDisplay: profile.label || '',
                };
            });
```

return {element:, value: , valueToDisplay}

src/components/GmChip.js
jparez: is not the member but the set (set value)

src/components/GmChip.js
jparez: Is a best pratice to (Property Reflection) in DOM we will have `<gm-chip type="error">`, and is usefull to target with CSS (gm-chip[type="error"])

src/components/GmTextInput.js
jparez: - regex-filter controls WHAT you can type (characters).
- strict-range controls HOW MUCH you can value (magnitude).
- regex-valid checks if the RESULT is correct (completeness).
All three are actively used in the project (e.g., in 
`Phone.js`,`Identifiers.js`,`GPS.js`,`Battery.js`)

src/components/GmTextInput.js
jparez: Currently, gm-text-input-change violates W3C standards by firing on every keystroke (like an input event) instead of only on commit (blur/enter). However, since many plugins (Phone, GPS, etc.) rely on this non-standard behavior for real-time validation, we cannot simply remove or change it without breaking them. We will refactor this in a dedicated commit to properly separate input (real-time) and change (commit) events and update all consumers accordingly.

src/components/GmTextInput.js
jparez: No, because #updateBottomVisibility() is also used by #updateUnitText(). Moving it inside #updateErrorMessage() would require duplicating the logic in #updateUnitText(). Keeping it as a shared helper prevents this duplication.

src/components/GmTextInput.js
jparez: regexFilter wasn't deleted

src/components/GmTextInput.js
jparez: the refacto of dipacthed event (input and change) will fix that

jparez: A separate commit fix the w3c compliant mistake (inversion of 'input' and 'change' events concept)

src/components/GmSlider.js
caarmen: OK, good to know!

src/components/GmChip.js
caarmen: Not sure why the thumbs down. It looks like you made the change?

src/components/GmDropdown.js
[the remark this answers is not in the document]
caarmen: Keeping innerHtml to be consistent with the rest of the project, and for now, no user data is provided in this component.

src/plugins/GAPPSInstall.js
caarmen: Thanks! Not really related to the component refactoring, but nice to have the fix anyway.

src/plugins/Network.js
caarmen: The old component's `setValue()` took an argument with a `.value` attribute, and accessed the value attribute. Now, the component takes the value directly as an argument, so the caller has to pass the value directly.

src/components/GmChip.js
caarmen: Maybe check if other components could benefit for `setAttribute`, in case some CSS selectors could be useful.

src/components/GmTextInput.js
caarmen: Ok. We initially thought it was only used for battery (to limit to 0-100, which is now managed by min/max bounds). But, in fact, this regex-filter is used for other rules (for example android id must be a hexadecimal string).

tests/unit/phone.test.js
caarmen: This case is different, because we're simulating typing text, not clicking an element.

src/plugins/Battery.js
caarmen: I don't see this anymore in the PR 🤷🏻

src/components/GmSwitch.js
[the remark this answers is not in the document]
caarmen: Fixed with a private variable `#boundHandleClick`

src/components/GmSlider.js
```diff
@@ -0,0 +1,221 @@
...
+                    this.input.min = this.#min;
+                    // Validate value against new min
+                    if (this.#value < this.#min) {
+                        this.setAttribute('value', this.#min);
+                    }
+                }
+                break;
+            case 'max':
+                // Validation for max: ensure it's not less than or equal to min
+                if (val <= this.#min) {
```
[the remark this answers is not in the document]
jparez: not sure this warning will be usefull, and why not trace all invalid keystroke ? On all invalid keystroke we log a warn ? this seems not serious

src/components/GmChip.js
jparez: The only other component that use attribute as CSS is gm-text-input and setAttribute is used.

src/components/GmChip.js
jparez: htis was a mistake, i had revert the change

caarmen (approved):
Cool to see this standardization of the different components! 🧩 

Huge thanks to you Jeremy for the time you invested in the pair review 🙏🏻

PR #171, opened 2026-03-16
[PLAYER-97] hide toolbar if no icons

jparez:
## Description

This PR changes the behavior of toolbars:

- An option has been added to display or hide the toolbar.

- If a toolbar (floating or legacy) has no elements, it is hidden (even if the "show" option is set to true).

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,7 +482,7 @@ A device renderer instance can be configured using the `options` argument (objec
 - **Type:** `Boolean`
 - **Default:** `true`
 - **Details:**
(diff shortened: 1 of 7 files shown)


pgivel (approved):
it would have been nice if the floating bar separator changes were in a separate commit, but LGTM anyway

jparez:
> it would have been nice if the floating bar separator changes were in a separate commit, but LGTM anyway

I admit I should have made atomic commit, sorry about that - note to the future

PR #174, opened 2026-03-18
[PLAYER-101] auto paste

jparez:
## Description

Introduces an "auto clipboard" feature that automatically sends the host system's clipboard content to the Android device when the user presses Ctrl+V (or Cmd+V on macOS) while the video element is focused and receive the copied text from device to the system's clipboard

### DIFF
--- a/src/plugins/Clipboard.js
+++ b/src/plugins/Clipboard.js
@@ -4,6 +4,21 @@ import '@/components/GmChip.js';
 import log from 'loglevel';
 log.setDefaultLevel('debug');
 
+// This is the moderne way (baseline 2020) to handle utf8 base64 conversion.
+const b64ToUtf8 = (str) => {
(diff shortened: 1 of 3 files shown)


src/plugins/Clipboard.js
```diff
@@ -142,8 +161,38 @@ export default class Clipboard extends OverlayPlugin {
     sendDataToInstance() {
         const json = {
             channel: 'framework',
-            messages: ['set_device_clipboard ' + window.btoa(unescape(encodeURIComponent(this.clipboardInput.value)))],
+            messages: ['set_device_clipboard ' + utf8ToB64(this.clipboardInput.value)],
         };
         this.instance.sendEvent(json);
     }
+
+    activeAutoClipboard() {
```
pgivel, line 169:
```suggestion
    activateAutoClipboard() {
```

pgivel (approved):
tiny typo

PR #176, opened 2026-03-24
[PLAYER-77] add new maps api to player

jparez:
## Description

gmaps marker is deprecated, this PR introduce the adding of the new google maps api to the player. Both (legacy and new) api are available

### DIFF
--- a/example/geny-window.html
+++ b/example/geny-window.html
@@ -18,7 +18,7 @@
          */
         // Legacy API: Direct Script Loading
         <script src="https://maps.googleapis.com/maps/api/js?key=YOURGOOGLEAPIKEY" async defer></script>
-        // Modern API: Dynamic Library Loading
+        // Modern API: Dynamic Library Loading (https://developers.google.com/maps/documentation/javascript/libraries)
         <script>
             (g=>{var h,a,k,p="The Google Maps JavaScript API",c="google",l="importLibrary",q="__ib__",m=document,b=window;b=b[c]||(b[c]={});var d=b.maps||(b.maps={}),r=new Set,e=new URLSearchParams,u=()=>h||(h=new Promise(async(f,n)=>{await (a=m.createElement("script"));e.set("libraries",[...r]+"");for(k in g)e.set(k.replace(/[A-Z]/g,t=>"_"+t[0].toLowerCase()),g[k]);e.set("callback",c+".maps."+q);a.src=`https://maps.${c}apis.com/maps/api/js?`+e;d[q]=f;a.onerror=()=>h=n(Error(p+" could not load."));a.nonce=m.querySelector("script[nonce]")?.nonce||"";m.head.append(a)}));d[l]?console.warn(p+" only loads once. Ignoring:",g):d[l]=(f,...n)=>r.add(f)&&u().then(()=>d[l](f,...n))})({
                 key: "YOURGOOGLEAPIKEY",
--- a/example/geny-window.html
+++ b/example/geny-window.html
@@ -12,8 +12,19 @@
         <script src="../dist/js/device-renderer.min.js" data-player-url></script>
 
         <!-- 
-        //to use maps option in gps widget you need to load google maps api and forward you google pai key
+        /*
+         * to use maps option in gps widget you need to load google maps api and forward you google api key
+         * Both google maps api (legacy : Direct Script Loading and modern : Dynamic Library Loading) are supported, use one of them and make sure to replace YOURGOOGLEAPIKEY with your actual API key.
+         */
+        // Legacy API: Direct Script Loading
         <script src="https://maps.googleapis.com/maps/api/js?key=YOURGOOGLEAPIKEY" async defer></script>
+        // Modern API: Dynamic Library Loading
+        <script>
+            (g=>{var h,a,k,p="The Google Maps JavaScript API",c="google",l="importLibrary",q="__ib__",m=document,b=window;b=b[c]||(b[c]={});var d=b.maps||(b.maps={}),r=new Set,e=new URLSearchParams,u=()=>h||(h=new Promise(async(f,n)=>{await (a=m.createElement("script"));e.set("libraries",[...r]+"");for(k in g)e.set(k.replace(/[A-Z]/g,t=>"_"+t[0].toLowerCase()),g[k]);e.set("callback",c+".maps."+q);a.src=`https://maps.${c}apis.com/maps/api/js?`+e;d[q]=f;a.onerror=()=>h=n(Error(p+" could not load."));a.nonce=m.querySelector("script[nonce]")?.nonce||"";m.head.append(a)}));d[l]?console.warn(p+" only loads once. Ignoring:",g):d[l]=(f,...n)=>r.add(f)&&u().then(()=>d[l](f,...n))})({
+                key: "YOURGOOGLEAPIKEY",
+                v: "weekly",
+            });
+        </script>
         -->
     </head>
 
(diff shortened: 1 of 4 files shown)


example/geny-window.html
```diff
@@ -12,8 +12,19 @@
...
 
         <!-- 
-        //to use maps option in gps widget you need to load google maps api and forward you google pai key
+        /*
+         * to use maps option in gps widget you need to load google maps api and forward you google api key
+         * Both google maps api (legacy : Direct Script Loading and modern : Dynamic Library Loading) are supported, use one of them and make sure to replace YOURGOOGLEAPIKEY with your actual API key.
+         */
+        // Legacy API: Direct Script Loading
         <script src="https://maps.googleapis.com/maps/api/js?key=YOURGOOGLEAPIKEY" async defer></script>
+        // Modern API: Dynamic Library Loading
```
caarmen: Perhaps a link to the google maps doc could be helpful to understand where this code comes from: https://developers.google.com/maps/documentation/javascript/libraries

src/plugins/GPS.js
```diff
@@ -485,11 +491,7 @@ export default class GPS extends OverlayPlugin {
     async checkLocationPermission() {
         try {
             this.permissionStatus = await navigator.permissions.query({name: 'geolocation'});
-            if (this.permissionStatus.state === 'denied') {
-                this.setToMyPositionBtn.disabled = true;
-            } else {
-                this.setToMyPositionBtn.disabled = false;
-            }
+            this.setToMyPositionBtn.disabled = this.permissionStatus.state === 'denied';
```
caarmen, line 498: Could be a bonus, for the review, for the future, to have separate commits:
* a commit for little cleanups like this (simplifying this if/else into a ternary operator expression)
* a separate commit for the new behavior (supporting the modern maps api).

src/plugins/GPS.js
```diff
@@ -698,8 +791,35 @@ export default class GPS extends OverlayPlugin {
...
+    destroy() {
+        this.clearMarkers();
+
+        if (this.map) {
+            google.maps.event.clearInstanceListeners(this.map);
+            this.map = null;
+        }
+
+        if (this.permissionStatus) {
+            this.instance.removeListener(this.permissionStatus, 'change');
```
caarmen: Where is `removeListener` defined?

Is `instance` a `DeviceRenderer`? I don't see `removeListener` defined on `DeviceRenderer` 🤔 Did I miss something?

src/plugins/GPS.js
```diff
@@ -636,24 +704,30 @@ export default class GPS extends OverlayPlugin {
...
+                position: {lat: numLat, lng: numLng},
+                map: this.map,
+                content: pinElement ? pinElement.element : null,
+            });
+        } else {
+            marker = new google.maps.Marker({
+                position: { lat: numLat, lng: numLng },
+                map: this.map,
+            });
+        }
```
caarmen, line 734: If I understand right, if we're in modern mode, but NOT with `AdvancedMarkerElement`, we'll create a legacy marker, right?

In `clearMarkers`, we do this:
```javascript
        this.markers.forEach((marker) => {
            if (this.isLegacyMode) {
                marker.setMap(null);
            } else {
                marker.map = null;
            }
```

For the cleanup, we only check `isLegacyMode`. We don't chaeck `AdvancedMarkerElement`.

What will happen in the cleanup in this case? It looks like we'll do `marker.map = null`. Is this right? Will this work on a legacy marker?

src/plugins/GPS.js
```diff
@@ -569,16 +571,36 @@ export default class GPS extends OverlayPlugin {
...
+
+                if (!this.isLegacyMode) {
+                    try {
+                        const response = await this.elevationService.getElevationForLocations(request);
+                        elevationResults = response.results;
+                    } catch (e) {
+                        log.error('Modern elevation request failed', e);
+                    }
+                } else {
+                    elevationResults = await new Promise((resolve) => {
```
caarmen, line 592: It looks like the way we handle elevation results for legacy is different than how we were handling this previously in the `main` branch.

Is this:
* part of the support for the modern api?
* or a separate cleanup to make the legacy elevation calculation more robust?

src/plugins/GPS.js
```diff
@@ -588,45 +610,91 @@ export default class GPS extends OverlayPlugin {
...
+                }
+                if (imports.elevation && imports.elevation.ElevationService) {
+                    ElevationServiceConstructor = imports.elevation.ElevationService;
+                }
+                if (imports.marker) {
+                    this.AdvancedMarkerElement = imports.marker.AdvancedMarkerElement;
+                    this.PinElement = imports.marker.PinElement;
+                }
+            } catch (e) {
+                log.warn('Failed to load Google Maps libraries dynamically', e);
```
caarmen, line 682: If we fail to import the modern imports, maybe other parts of the code which do the following will fail:
```javascript
if(!this.isLegacyMode) {
    // Do stuff with some of the modern apis
}
```

Should we perhaps force `this.isLegacyMode` to `true` inside this `catch(e)`? And rollback some of the constructors we assigned inside the `try`?

I imagine it could be difficult to test this error case.

src/plugins/GPS.js
```diff
@@ -666,27 +740,46 @@ export default class GPS extends OverlayPlugin {
...
-                {
-                    locations: [location],
-                },
-                (results, status) => {
-                    if (status === 'OK' && results && results[0]) {
-                        const currentAltitude = Number(this.inputComponents.altitude.value);
-                        const newElevation = Number(results[0].elevation);
-                        if (currentAltitude !== newElevation) {
-                            this.setFieldValue('altitude', newElevation);
+            const LatLngClass = this.isLegacyMode ? google.maps.LatLng : (this.LatLng || google.maps.LatLng);
```
caarmen, line 747: What will happen if we're in modern mode (`this.isLegacyMode` is `false`) but we don't yet have `this.LatLng`?

I guess this could happen if `addMapMarker()` is called before `loadMap()`. Probably not a realistic case.

But if that happens, we'll be passing a legacy `google.maps.LatLng` instance to the modern `getElevationForLocations()`. Not sure if that's a problem.

src/plugins/GPS.js
caarmen: I just noticed that this block (we're in `getLocation()` here) is essentially copy/paste of a similar block in `addMapMarker()`.

Before (`main` branch): The logic to calculate elevation was different in `getLocation()` vs `addMapMarker()`.
Now: the logic seems the same in both places.

Did I read this right?

src/plugins/GPS.js
jparez: big fail... addListener return a remove cb fn... there is no removeListener method on instance

src/plugins/GPS.js
jparez: is a little dirty i admit but both fn take same param so isn't a problem

src/plugins/GPS.js
jparez: But it depends on client integration. If client try to load with modern api, no chance that code for load lagcy api exist

src/plugins/GPS.js
jparez: yes, code has evolve for modern api support (return a promise)

src/plugins/GPS.js
jparez: for robust clean check we can do (marker && typeof marker.setMap === 'function')

src/plugins/GPS.js
caarmen: I see it's fixed now!

caarmen (approved):

PR #178, opened 2026-03-27
[Player-72]  implement animated orientation change 

jparez:
## Description

This PR improves the device rotation UX and fixes layout sizing so the player container always follows the rendered video size.

What changed
- Added smooth orientation transition animation when rotating the device.
- Updated orientation detection using video dimension changes (ResizeObserver).
- Applied orientation/form-factor classes (portrait/landscape, phone/tablet) on the video wrapper.
- Reworked phone border rendering in CSS and **removed manual JS border repositioning logic**.


src/plugins/ButtonsEvents.js
```diff
@@ -29,9 +29,48 @@ export default class ButtonsEvents {
...
 
         if (this.instance.options.rotation) {
             this.registerToolbarButton('ROTATE', ROTATE_KEYCODE, 'gm-rotation', i18n.BUTTONS_ROTATE || 'Rotate', false);
+
+            // animate orientation change based on video dim change
+            let lastMode = null;
+            let lastWidth = null;
+            let firstRun = true;
+            this.videoResizeObserver = new ResizeObserver(async (entries) => {
+                const mode = entries[0].target.videoWidth > entries[0].target.videoHeight ? 'landscape' : 'portrait';
```
pgivel: Maybe put a guard on entries[0] ?
Like this?

```suggestion
            this.videoResizeObserver = new ResizeObserver(async (entries) => {
                let video = entries[0].target;
                if (video != this.instance.video) {
                  return;
                }
                const mode = video.videoWidth > video.videoHeight ? 'landscape' : 'portrait';
```
etc?

src/plugins/ButtonsEvents.js
```diff
@@ -66,6 +105,18 @@ export default class ButtonsEvents {
...
             this.registerToolbarButton('POWER', POWER_KEYCODE, 'gm-power', i18n.BUTTONS_POWER || 'Power', true);
         }
+
+        this.instance.registerEventCallback('sensors', (message) => {
+            const values = message.split(' ');
+            // Message format: "accelerometer <x> <y> <z>"
+            if (values.length !== 3 || values[1] !== 'accelerometer') {
+                return;
+            }
+            const [x, y] = values[2].split(':');
```
pgivel, line 119: This doesn't match the "Message format" comment right above, it doesn't have ":" characters. Maybe the comment should be something like this? `// Message format: "accelerometer <x>:<y>:<z>"`

src/plugins/ButtonsEvents.js
```diff
@@ -185,4 +236,58 @@ export default class ButtonsEvents {
...
+         * to match the previous state, then animate the return to 0 degrees.
+         */
+        if (mode === 'portrait' && formFactor === 'phone' || mode === 'landscape' && formFactor === 'tablet') {
+            container.style.transform = 'rotate(-90deg) scale('+scale+')';
+        } else {
+            container.style.transform = 'rotate(90deg) scale('+scale+')';
+        }
+
+        // Force a synchronous layout reflow to ensure the 'transition: none' and initial transform are applied before the animation starts.
+        void container.offsetHeight;
```
pgivel, line 272: I don't understand this line. Does it do anything?

src/plugins/ButtonsEvents.js
```diff
@@ -185,4 +236,58 @@ export default class ButtonsEvents {
...
+
+    async triggerOrientationAnimation(mode, formFactor, scale) {
+        const container = this.instance.videoWrapper;
+        const playerScreenWrapper = this.instance.playerScreenWrapper;
+
+        // Avoid any scrollbar display during the animation by hiding overflow on the wrapper
+        playerScreenWrapper.style.overflow = 'hidden';
+        // Reset transition to prepare for the immediate "counter-rotation"
+        container.style.transition = 'none';
+        container.style.transform ='';
```
pgivel: Maybe these line could be moved after the multi-line comment that explains that we'll do an unanimated "counter-rotation" first. Functionnaly the same, but easier to read and understand.

src/plugins/ButtonsEvents.js
```diff
@@ -185,4 +236,58 @@ export default class ButtonsEvents {
...
+                if (event && event.target) {
+                    if (event.target !== container || event.propertyName !== 'transform') {
+                        return;
+                    };
+                }
+
+                // Clean up styles
+                container.style.transition = '';
+                container.style.transform = '';
+                playerScreenWrapper.style.overflow = '';
```
pgivel: This feels a bit redundant with lines 247 & 248 where we force-remove the transition&transform. And also if there were any transitions & transform before this function, it wipes them. (same for overflow).

Maybe right before lines 247 & 248 we could read the style properties, and reapply them here? Or maybe they're supposed to be set via CSS anyway, not via html style properties? So then simply removing either the reset here or the wipe at lines 247 & 248 ?

pgivel (changes requested):
This feels like much cleaner code, and nice behaviour :) A few comments in the resizeobeserver handler.

src/plugins/ButtonsEvents.js
jparez: Yes, this is a "synchronous layout flush". Without it, the browser's optimization engine could batch the two transform changes together, skipping the initial state and jumping directly to the end of the animation.

By accessing offsetHeight, we force the browser to calculate the styles now, effectively 'recording' the first state (the rotation) before we trigger the transition to the second state.

check https://gist.github.com/paulirish/5d52fb081b3570c81e3a to get a full list of commands which force a reflow

pgivel: TIL

src/plugins/ButtonsEvents.js
jparez: In line 247, we force transition to none, which is different from setting it to ''. Using '' only clears the inline style, whereas none explicitly ensures no transition is applied (even if one is defined in the CSS). However, you're right—saving the existing inline styles is a better approach for a reusable component to avoid overwriting a user's custom settings.

pgivel (approved):

PR #179, opened 2026-04-01
[PLAYER-103] refacto default turn icon and tooltip

jparez:
## Description

refacto default turn icon and tooltip

Tested on a PaaS device (injected the built .js and .css files).

without connectionFailedURL option : 
image

with connectionFailedURL option : 

image


src/plugins/util/TooltipManager.js
```diff
@@ -33,15 +33,17 @@ export default class TooltipManager {
      * @param {HTMLElement} element - The target element
      * @param {string} text - The tooltip text
      * @param {string} [position] - 'top'|'bottom'|'left'|'right' - Optional. Preferred position for the element. If not specified, the position will be determined in the following order of preference: 'bottom', 'left', 'right', 'top'.
-     *  @param {string} [classes] - Optional. Additional classes to add to the tooltip.
+     * @param {string} [classes] - Optional. Additional classes to add to the tooltip.
+     * @param {boolean} [isHTML] - Optional. If true, the tooltip text will be treated as HTML. BEWARE: This can lead to XSS vulnerabilities if the text is not sanitized or injected by users
      */
-    setTooltip(element, text, position, classes = null) {
+    setTooltip(element, text, position, classes = null, isHTML = false) {
+        // isHTML is used to display HTML content in the tooltip. It should be used with caution, especially if the text content is user-generated (translation, ...).
```
pgivel: This comment feels a bit redundant with the docstring right above

pgivel (approved):
small comment, LGTM otherwise

src/plugins/util/TooltipManager.js
jparez: True, but I'm not sure developers will read the JSDoc, especially with a clearly named param. I prefer to keep the warning, but i can merge the 'beware' from the jsdoc into this sentence and clean up the jsdoc. Let me know what you think

pgivel: as you wish, I don't have a strong opinion on this :)

pgivel (approved):

src/plugins/util/TooltipManager.js
jparez: After reflection, I prefer to keep both. Currently, the JSDoc isn't very useful because objects are marked as 'any', but I'd like to add TypeScript to the project later. In that case, JSDoc will be very helpful. This way, we’ll have a visual indicator (like an alert emoji) on hover, as well as the warning inside the function code for future updates.

PR #180, opened 2026-04-22
fix epic issues

jparez:
## Description

- fix gapss text on isntall section
- fix error on gps plugin while widget is destroying
- fix toolbar doen'st show if floating bar is hidden
- fix css integration for desktop
- fix loop on autopaste with arm => refacto feature

## Type of change

### DIFF
--- a/src/plugins/GPS.js
+++ b/src/plugins/GPS.js
@@ -819,7 +819,7 @@ export default class GPS extends OverlayPlugin {
             this.removePermissionStatusListener = null;
         }
 
-        if (this.removeMapClickListener) {
+        if (this.removeMapClickListener && this.removeMapClickListener === 'function') {
             this.removeMapClickListener();
             this.removeMapClickListener = null;
         }
@@ -829,7 +829,5 @@ export default class GPS extends OverlayPlugin {
         }
 
         delete this.instance.gps;
-
-        super.destroy();
     }
 }
--- a/src/plugins/GPS.js
+++ b/src/plugins/GPS.js
@@ -833,8 +833,14 @@ export default class GPS extends OverlayPlugin {
             this.removePermissionStatusListener = null;
         }
 
-        if (this.removeMapClickListener && this.removeMapClickListener === 'function') {
-            this.removeMapClickListener();
+        if (this.removeMapClickListener) {
+            if (typeof this.removeMapClickListener.remove === 'function') {
+                this.removeMapClickListener.remove();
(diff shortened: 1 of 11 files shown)


src/plugins/GPS.js
```diff
@@ -819,7 +819,7 @@ export default class GPS extends OverlayPlugin {
             this.removePermissionStatusListener = null;
         }
 
-        if (this.removeMapClickListener) {
+        if (this.removeMapClickListener && this.removeMapClickListener === 'function') {
```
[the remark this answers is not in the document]
pgivel: the last part of this command is unrelated, but the first part is true I think.

src/plugins/KeyboardEvents.js
```diff
@@ -31,9 +33,9 @@ const INVISIBLE_KEYS = {
 };
 
 const CTRL_SHORTCUT_KEYS = {
-    67: '0x010000cf', // Copy
-    86: '0x010000e2', // Paste
-    88: '0x010000d0', // Cut
+    c: '0x010000cf', // Copy (legacy keyCode: 67)
+    v: '0x010000e2', // Paste (legacy keyCode: 86)
```
pgivel: Shouldn't this 0x010000e2 value be factorized between KeyboardEvents and Clipboard?

src/plugins/KeyboardEvents.js
```diff
@@ -168,7 +185,6 @@ export default class KeyboardEvents {
         this.currentlyPressedKeys.set(key, key);
         event.preventDefault();
         event.stopPropagation();
-        event.returnValue = false;
         return false;
```
pgivel: With the pile of if/else it's, most with return statements, it's starting to get not obvious which cases trickle down to here. But I'm not sure how we could improve the readability of this code. (same in onKeyUp handler, although it's a bit more compact and readable there).

pgivel (changes requested):

src/plugins/GPS.js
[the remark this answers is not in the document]
jparez: both comments are correct

src/plugins/KeyboardEvents.js
jparez: obviously

src/plugins/KeyboardEvents.js
jparez: That's why i added some comments before each 'if' statements. Maybe a swicth with an array of json could be more readable. let's try this

pgivel (approved):

PR #182, opened 2026-04-30
Release 4.3.1

jparez:
## Description

Release 4.3.1 with OIDC connect to NPM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ on:
 
 permissions:
     id-token: write
-    contents: write
+    contents: read
 
 jobs:
     release:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
             - uses: actions/checkout@v6
             - uses: actions/setup-node@v6
               with:
-                  node-version: 20
+                  node-version: 24
```
caarmen, line 23: https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow

Looks like 22 is the minimum required.

I guess it doesn't matter though, it doesn't impact the build of our package, right?

